PR #137, opened 2013-04-23
Support rotation when tracing to new logfile

gomoripeti:
lager:trace_file/3 can now also handle a new-style configuration proplist
to allow full configuration of a newly created file backend.
If the file backend already exists, however, only the level can be given
(similarly to the old behaviour).

I initially created this commit weeks ago. It would be nice to have persistent tracing into file fully configurable (and I am ready to add this to this branch) but I am not sure how the syntax should look like. (My concern is the handling of Level: should it be part of the backend config or is it a separate mandatory argument of the tracer)

### DIFF
--- a/src/lager.erl
+++ b/src/lager.erl
@@ -143,29 +143,50 @@ log(Level, Metadata, Format, Args) when is_list(Metadata) ->
 trace_file(File, Filter) ->
     trace_file(File, Filter, debug).
 
-trace_file(File, Filter, Level) ->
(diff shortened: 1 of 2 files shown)


mrallen1:
This PR is pretty old so I'm not sure of its status. @Vagabond do you recall anything about this PR?

mrallen1:
Haven't heard anything on this for a year. I'm going to close it. Please reopen if needed.

PR #208, opened 2014-02-27
Update rebar.config

MisaKondo:
A things I added for my own use, you can have them if you want.


Vagabond:
Do not want, if this is a problem put something like this inf your .gitconfig:

[url "http://github.com"]
  insteadOf = ssh://git@github.com/

PR #227, opened 2014-07-28
Fix specification of backend config in log rotation example

derek121:
The example in "Internal log rotation" includes the tuple
{name, "error.log"}

This apparently should be

{file, "error.log"}

as seen in other config examples in README.md.


Vagabond:
+1

PR #233, opened 2014-09-15
Mention the color atom for the formatter config

dvaergiller:
The readme did not mention how to set color when using custom formatting. This "fixes" that.


Vagabond:
+1

PR #241, opened 2014-11-21
Add optional log_root environment variable

stolen:
When using multiple log files, usually all filenames have common directory.
When application has several environments (e.g. development and production) the log directory may differ.
When adding a new log file, the developer has to be careful to edit log path according to the environment.

Moving log directory to the separate environment variable allows to apply logging changes by using just ```diff``` and ```patch``` utilities. Also application config becomes shorter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ your app.config):
 
 ```erlang
 {lager, [
+  {log_root, "/var/log/hello"},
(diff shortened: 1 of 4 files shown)


Vagabond:
+1

AntonSizov:
+1

PR #246, opened 2014-12-11
add per-module per-function trace example

lrascao:
was not documented, very useful in production scenarios where default debug level is critical


Vagabond:
+1

PR #247, opened 2014-12-17
make trace and stop_trace take the same arguments

lrascao:
starting a filtered trace is easy enough, one would expect the stop_trace to receive the same arguments.
for example, starting a trace:
lager:trace(lager_console_backend, [{module, users_serv}, {function, is_client_update_available}], debug).

to stop the trace, one must issue the following (non-intuitive) command:
lager:stop_trace({{all,[{module,'=',users_serv},{function,'=',is_client_update_available}]},{mask,255},lager_console_backend}).

with this PR the stop_trace now becomes:
lager:stop_trace(lager_console_backend, [{module, users_serv}, {function, is_client_update_available}], debug).

### DIFF
--- a/src/lager.erl
+++ b/src/lager.erl
@@ -27,7 +27,7 @@
         log/3, log/4,
         md/0, md/1,
         trace/2, trace/3, trace_file/2, trace_file/3, trace_file/4, trace_console/1, trace_console/2,
-        clear_all_traces/0, stop_trace/1, status/0, 
+        clear_all_traces/0, stop_trace/3, status/0, 
         get_loglevel/1, set_loglevel/2, set_loglevel/3, get_loglevels/0,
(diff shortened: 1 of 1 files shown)


Vagabond:
This breaks one of the tests, it is documented that you can pass the result of start_trace into stop trace, so you should fix start_trace to return the 3-tuple. This is as simple as changing line 193 of lager.erl to return {ok, Trace0} rather than {ok, Trace}.

I think this is fine once that change is made.

Vagabond:
Ah, you also need to export stop_trace/1 as well, to make the tests pass.

Vagabond:
ah, that isn't right either, let me just rework this and make a new pull request.

Vagabond:
This is superceded by #251

lrascao:
closing, #251 should be focused on instead

PR #248, opened 2015-01-05
Add lager_manager_killer and tests for it.

jinnipark:
Changing async from true to false under lager overload is just putting the problem from one place to another.  It actually blocks all the log producers that they - deprived of the chance to offload - are killed helpless.  IMHO,  there is no way but to drop delayed logs for the entire system to stay healthy.  I chose to kill the event manager and wait for all the handlers being reinstalled by the watchers.

### DIFF
--- a/src/lager_manager_killer.erl
+++ b/src/lager_manager_killer.erl
@@ -1,0 +1,44 @@
+-module(lager_manager_killer).
+-author("Sungjin Park <jinni.park@gmail.com>").
+-behavior(gen_event).
+
+-export([init/1, handle_event/2, handle_call/2, handle_info/2, terminate/2, code_change/3]).
+
+-include("lager.hrl").
+
+-record(state, {
+  killer_hwm :: non_neg_integer(),
+  killer_reinstall_after :: non_neg_integer()
+}).
+
+init([KillerHWM, KillerReinstallAfter]) ->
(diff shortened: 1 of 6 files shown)


jonmeredith:
@Vagabond any thoughts on this?

src/lager_manager_killer.erl
```diff
@@ -0,0 +1,44 @@
...
+  {ok, {mask, ?LOG_NONE}, State};
+handle_call({set_loglevel, _Level}, State) ->
+  {ok, ok, State};
+handle_call(get_settings, State = #state{killer_hwm=KillerHWM, killer_reinstall_after=KillerReinstallAfter}) ->
+  {ok, [KillerHWM, KillerReinstallAfter], State};
+handle_call(_Request, State) ->
+  {ok, ok, State}.
+
+handle_event({log, _Message}, State = #state{killer_hwm=KillerHWM, killer_reinstall_after=KillerReinstallAfter}) ->
+  {message_queue_len, Len} = process_info(self(), message_queue_len),
```
jonmeredith: is driving the kill decision off this local message queue len going to have the desired behavior on all messages?

mrallen1:
I am not opposed to the idea, just not sure of the full ramifications of this measurement/decision point.

jinnipark:
Updated in https://github.com/basho/lager/pull/328

PR #249, opened 2015-01-20
Flow control messages to lager_file_backend by adding high water mark

ericliang:
Services using lager have to face high velocity of incoming messages, which would overflow the file backend. The situation is desperate when something abnormal events happen, and especially we're using cloud hosts.

The config error_logger_hwm only effects error_logger redirected but the raw file backend,  IIRC. So I add a similar configuration high_water_mark.

To be friendly for the PR, I've split it to three parts: 
1. Refactor the error_logger_hwm control codes, make it public to all modules by moving to lager_util;
2. add high_water_mark configuration to lager_file_backend;
3. add set_loghwm method to change high water mark on the fly.

Thanks in advance for your review.

### DIFF
--- a/include/lager.hrl
+++ b/include/lager.hrl
@@ -100,3 +100,15 @@
     end)).
 -endif.
 
+-record(lager_shaper, {
+		  %% how many messages per second we try to deliver
+		  hwm = undefined :: 'undefined' | pos_integer(),
+		  %% how many messages we've received this second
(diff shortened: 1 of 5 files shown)


mrallen1:
Hey thanks for the patches - we are considering this for inclusion in an upcoming release.  Sorry for the long delay.

mrallen1:
Just pushed b4ab519 which implements your PR in a rebased way for 3.0

PR #250, opened 2015-01-22
add application stop/start suppression for ct logs

evanmcc:
in common_test, starting and stopping applications can make your logs
very noisy.  the patch allows these messages to be suppressed on the
happy path, and raises the high water mark during ct tests so that
messages aren't dropped.


seancribbs:
:+1: df3cf16

Vagabond:
+1 on this as well, verified it works as intended.

seancribbs:
@borshop merge

PR #252, opened 2015-01-28
Fix info in README on metadata fallbacks in formatters' configs

mentels:
The third element has to be an "semi-iolist".


michalwski:
:+1:

PR #253, opened 2015-02-11
Strict config

haguenau:
This patch makes it so that handler config entries with the wrong number of values (i.e., forms other than the two expected members "module" and "config") are rejected at startup, rather than being silently ignored (since list comprehensions just drop any failed pattern matches).

Here is an example of a bad handler that would previously have been silently rejected:

      {lager_logstash_backend,
        {level, info},
        {output, {udp, "localhost", 5000}},
        {format, json},
        {json_encoder, jiffy}
      }

Here is what was actually meant:

      {lager_logstash_backend, [
        {level, info},
        {output, {udp, "localhost", 5000}},
        {format, json},
        {json_encoder, jiffy}
      ]}

(Note how it's a two-element tuple, where the second member is an association list, whereas the first example is a copy-pasting error.)

### DIFF
--- a/src/lager_app.erl
+++ b/src/lager_app.erl
@@ -113,7 +113,7 @@ start(_StartType, _StartArgs) ->
                 end
         end,
 
-    _ = lager_util:trace_filter(none), 
+    _ = lager_util:trace_filter(none),
 
     {ok, Pid, SavedHandlers}.
 
@@ -153,7 +153,7 @@ add_configured_traces() ->
 
 maybe_make_handler_id(Mod, Config) ->
     %% Allow the backend to generate a gen_event handler id, if it wants to.
-    %% We don't use erlang:function_exported here because that requires the module 
+    %% We don't use erlang:function_exported here because that requires the module
     %% already be loaded, which is unlikely at this phase of startup. Using code:load
     %% caused undesireable side-effects with generating code-coverage reports.
     try Mod:config_to_id(Config) of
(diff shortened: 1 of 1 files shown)


Vagabond:
Would it be better to log the bad configs rather than just function_clause erroring? Can you add a test for this?

haguenau:
* Andrew Thompson <notifications@github.com>, 2015-02-12 14:57:29 Thu:
> Would it be better to log the bad configs rather than just
> function_clause erroring? Can you add a test for this?

Both your suggestions make sense; I'll have a look when time permits.

mrallen1:
@jonmeredith @macintux We should add this for a 3.0 release - would be a good catch to debug configs especially since we're adding new stuff for multi-sink.

mrallen1:
In the multiple-sinks branch, we completely changed how handlers are started. They are now iterated over in a map which has the same failure semantics as this PR, but some additional error checking tests would be a good idea.

mrallen1:
Just committed 31cf6c0 which addresses this concern, but implements it in a different way.

mrallen1:
31cf6c0

haguenau:
> Just committed 31cf6c0 which addresses this concern, but implements it in a different way.

Oh, hey, that's great. Good to see this fixed.

mrallen1:
A version was merged into 3.0 and is now on master. Closing.

haguenau:
> A version was merged into 3.0 and is now on master. 

Thanks!

PR #254, opened 2015-02-12
Improve format_reason for certain oddly formed badarg stacktraces

Vagabond:
I'm not sure how these weird stacktraces are formed, but @jaredmorrow found one, so here's a band-aid.


jaredmorrow:
+1, gives a more sane output for the badarg failure we saw

PR #255, opened 2015-03-13
Provide means to compress printed record output

keynslug:
Sometimes there are a LOT of records going through logging backends
which contain notable number of empty (i.e. undefined) fields. This
humble patch solemnly introduces new function `lager:pr/3` designed to
help cope with such kind of issues, providing a one and only option `compress`
which, when turned on, effectively eliminates any empty fields in records.


mrallen1:
Good candidate for 3.0 @jonmeredith @macintux

mrallen1:
Rebased / merged in `intg` - all tests pass (after minor fix up)

PR #258, opened 2015-04-15
trace_file don't work, if env "log_root" is set.

MaximMinin:
This is a little bugfix for trace_file funktion.


mrallen1:
This should get into 3.0, if not a 2.x point release @jonmeredith @macintux

mrallen1:
Rebased and merged into intg => all tests pass

mrallen1:
Also going to put this on `mra/otp-17` as it seems like a good bug fix for 2.x

PR #259, opened 2015-04-22
Fix  FORMAT ERROR in error_logger backend

mkurkov:
Hi,
I stumbled upon errors in log file about FORMAT ERROR, it turns out that problem was that some library sometimes sends unicode data via error_logger:xxx_msg calls. 
Looking into code I saw that problem is in error_logger handler that call lager:safe_format twitce in some cases - first properly with Fmt, Args, and second time with already formatted string (in lager:log).
As no unicode characters are allowed in format string I see this errors in log.
So I removed  premature formating from error_logger backend for messages, calling just lager:log, it seems that solves the problem.

### DIFF
--- a/src/error_logger_lager_h.erl
+++ b/src/error_logger_lager_h.erl
@@ -197,7 +197,7 @@ log_event(Event, State) ->
                     ?LOGFMT(error, Pid, "Webmachine error at path ~p : ~s", [Path, format_reason(StackTrace)]);
                 _ ->
                     ?CRASH_LOG(Event),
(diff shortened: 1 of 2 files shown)


mrallen1:
I think this looks good. We need to test it before it gets merged, of course. @macintux what do you think?

macintux:
Agreed on both points, thanks @mkurkov. I'll take a closer look soon.

tburghart:
These changes, adjusted for the strategy we've chosen for handling extra message data, have been merged into the "feature/otp-17" branch.

mkurkov:
@tburghart Great! When do you plan to merge this in the mainline, so we will be able to move from our fork? Speaking of unicode support your syslog_lager backend definitely need some work on it.

Issue #223, opened 2014-07-17
Lager does not perform any group leader redirection from error_logger messages [JIRA: RIAK-1897]


josevalim:
By default, Erlang keeps an `error_logger` handler running for the sole purpose of redirecting messages between nodes:

https://github.com/erlang/otp/blob/OTP-17.0/lib/kernel/src/error_logger.erl#L367-L370

You can use `gen_event:which_handlers(error_logger)` to see there is always an `error_logger` handler in there. All other erlang handlers by default ignore messages coming from other nodes as they assume a redirection will happen, for example:

https://github.com/erlang/otp/blob/OTP-17.0/lib/stdlib/src/error_logger_tty_h.erl#L54-L55

However, lager by default removes all error_logger handlers:

https://github.com/basho/lager/blob/2.0.3/src/lager_app.erl#L109-L110

And the default logger installed by lager does not do any redirection based on the node group leader:

https://github.com/basho/lager/blob/2.0.3/src/error_logger_lager_h.erl#L148

This basically means that, after lager is installed, there is no node redirection happening. The main downside of this behaviour is that when using remote shells, the log messages are no longer redirected. The same with executing processes on remote nodes, where you may expect at the least the crash to be reported in the caller.

However, lager may have good reasons to remove this behaviour, however I could not find a rationale anywhere. So is it a bug or a feature? :D

mrallen1:
In 3.x we have group leader support. Still probably need to fix up the emulator messages though.

mrallen1:
Closing this for now, please reopen if we missed something.

PR #263, opened 2015-05-04
Add support for group leader strategies

DeadZen:
The default strategy is 'handle', for backwards compatibility,
and is set to that if otherwise undefined.  Other strategies
include 'mirror' and 'ignore'.
   Before when the error_logger was removed when lager starts up,
some functionality was chopped off, such as forwarding io to the
caller's group leader, this is supported with the strategy 'ignore'
and is similar to how the io system in erlang works traditionally.
   Setting the 'mirror' strategy will essentially combine the other
two strategies such that the remote node reply is forwarded and also
handled locally. This is in some cases more in-line with what some
people might expect, efficiency may vary.
   The traditional 'ignore' approach is more suitable and most likely
originally intended for embedded terminals where 'mirror'-ing the io
is largely unnecessary.

### DIFF
--- a/src/error_logger_lager_h.erl
+++ b/src/error_logger_lager_h.erl
@@ -41,7 +41,9 @@
         %% the current second
         lasttime = os:timestamp() :: erlang:timestamp(),
         %% count of dropped messages this second
-        dropped = 0 :: non_neg_integer()
+        dropped = 0 :: non_neg_integer(),
+        %% group leader strategy
+        groupleader_strategy :: handle | ignore | mirror
     }).
 
 -define(LOGMSG(Level, Pid, Msg),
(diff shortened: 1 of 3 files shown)


DeadZen:
Additionally, this adds env key {error_logger_groupleader_strategy, handle | ignore | mirror} and addresse issues #223 and #225.

mjbrender:
Love seeing your knowledge drop @DeadZen :+1:

mrallen1:
@jonmeredith @macintux This is a good candidate for the 3.0 release we talked about.

mrallen1:
Just rebased into the integration branch; all tests pass

DeadZen:
Nice, thank you guys, btw I pushed 0.1.7 of goldrush last night, It completes the library logic (adds '>=' | '=<') operators and provides more for lager tracing. I could update the dependency here if that's alright.

src/error_logger_lager_h.erl
```diff
@@ -74,16 +76,24 @@ set_high_water(N) ->
...
+    {ok, #state{hwm=HighWaterMark, groupleader_strategy=GlStrategy}}.
 
 handle_call({set_high_water, N}, State) ->
     {ok, ok, State#state{hwm = N}};
 handle_call(_Request, State) ->
     {ok, unknown_call, State}.
 
-handle_event(Event, State) ->
+handle_event(Event, #state{groupleader_strategy=Strategy}=State) ->
+    GL = element(2, Event),
```
aerosol: :ok_hand:

mrallen1:
3.0 is merged on master now. Closing.

PR #264, opened 2015-05-13
Support multiple sinks in lager

mrallen1:
This is a feature which lays the ground work for an application to use and have multiple event sinks for use in lager.  This is useful in the scenario where a customer (for instance) wants to **always** log security or access events for audit purposes and is willing to pay the latency costs and assume the overload risks associated with that.

To support this feature, many portions of lager have been modified to accept a Sink argument, including the parse transform which dispatches log messages into gen_event. At the same time, this change should be totally backwards compatible.

Benchmarking runs show that lager's performance is not substantially different with or without multiple sinks enabled, so with this feature enabled, performance is no worse than pre-multiple-sink lager's performance.


mrallen1:
@Vagabond If you have time to review this, your input would be gratefully received.  Thanks.

README.md
```diff
@@ -16,6 +16,7 @@ Features
 * When no handler is consuming a log level (eg. debug) no event is sent
   to the log handler
 * Supports multiple backends, including console and file.
+* Supports multiple sinks (and eventually different sync/async policies per sink)
```
macintux: It supports different policies today, unless I'm misremembering

README.md
```diff
@@ -92,6 +93,65 @@ your app.config):
...
+there were an `audit` sink, we might be willing to pay the latency costs to log
+all of those messages to disk and we don't want lager to drop them in an
+overload scenario.)
+
+To use multiple sinks (beyond the built in sink of lager and lager_event), you
+need to:
+
+1. Setup rebar.config
+2. Configure the backends in app.config
+
```
macintux: We need to discuss naming here, too.

README.md
mrallen1: OK, let's get it fixed. :-)

macintux: Will update it.

src/lager.erl
```diff
@@ -126,6 +137,11 @@ do_log(Severity, Metadata, Format, Args, Size, SeverityAsInt, LevelThreshold, Tr
 dispatch_log(Severity, _Module, _Function, _Line, _Pid, Metadata, Format, Args, Size) ->
     dispatch_log(Severity, Metadata, Format, Args, Size).
 
+
+%% TODO:
```
macintux: Probably should decide what to do about this

README.md
```diff
@@ -16,6 +16,7 @@ Features
 * When no handler is consuming a log level (eg. debug) no event is sent
   to the log handler
 * Supports multiple backends, including console and file.
+* Supports multiple sinks (and eventually different sync/async policies per sink)
 * Rewrites common OTP error messages into more readable messages
 * Support for pretty printing records encountered at compile time
 * Tolerant in the face of large or many log messages, won't out of memory the node
```
macintux: We also need to discuss the impact on tracing.

README.md
mrallen1: Yeah we definitely do. It's confusing and deserves clarity if not for external developers' sake - but our own!

src/error_logger_lager_h.erl
```diff
@@ -74,7 +74,7 @@ set_high_water(N) ->
     gen_event:call(error_logger, ?MODULE, {set_high_water, N}, infinity).
 
 -spec init(any()) -> {ok, #state{}}.
-init([HighWaterMark]) ->
+init([{sink, Sink}, HighWaterMark]) ->
```
Vagabond: Should we be discarding this parameter? It also triggers a compiler warning.

Vagabond:
I consistently get a test failure:

```
in function lager_test_backend:'-test_body/2-fun-3-'/0 (test/lager_test_backend.erl, line 715)
in call from lager_test_backend:'-error_logger_redirect_crash_test_/0-fun-2-'/2 (test/lager_test_backend.erl, line 728)
**error:{assertion_failed,[{module,lager_test_backend},
                   {line,715},
                   {expression,"false"},
                   {expected,true},
                   {value,false}]}
```

Vagabond:
But that failure seems present on master as well, hmm.

macintux:
That's an OTP-17 thing. There's a patch or PR out there somewhere.

src/error_logger_lager_h.erl
macintux: Sink is only necessary for the throttle backend, so all of the other backends will ignore it. Not sure why I didn't name the binding appropriately there.

mrallen1:
I think @tburghart has a branch with the fix in it for OTP 17.  I meant to merge his branch into this one as an experiment but I never got to it.

src/lager_transform.erl
```diff
@@ -75,138 +77,148 @@ walk_clauses(Acc, [{clause, Line, Arguments, Guards, Body}|T]) ->
...
+    LevelVar = make_varname("__Level", Line),
+    TracesVar = make_varname("__Traces", Line),
+    PidVar = make_varname("__Pid", Line),
+    %% Wrap the call to lager:dispatch_log/6 in case that will avoid doing any work if this message is not elegible for logging
+    %% See lager.erl (lines 89-100) for lager:dispatch_log/6
+    %% case {whereis(Sink), whereis(?DEFAULT_SINK), lager_config:get({Sink, loglevel}, {?LOG_NONE, []})} of
+    {'case',Line,
+         {tuple,Line,
+                [{call,Line,{atom,Line,whereis},[{atom,Line,SinkName}]},
+                 {call,Line,{atom,Line,whereis},[{atom,Line,lager_event}]}, %% <--lager_event is ?DEFAULT_SINK
```
macintux: We include `lager.hrl` in this file; can we not just use `?DEFAULT_SINK`?

src/lager.erl
mrallen1: I hate the `log2` name, but I couldn't think of a better way to handle this. It feels like there **ought** to be a function with arity 4 that takes Sink, Level, Pid and Message. Open to naming suggestions here.

src/lager_transform.erl
```diff
@@ -75,138 +77,148 @@ walk_clauses(Acc, [{clause, Line, Arguments, Guards, Body}|T]) ->
...
+                                           end;
+                                       [Attrs, Format, Args] ->
+                                           {concat_lists(Attrs, DefaultAttrs), Format, Args}
+                                   end,
+    %% Generate some unique variable names so we don't accidentaly export from case clauses.
+    %% Note that these are not actual atoms, but the AST treats variable names as atoms.
+    LevelVar = make_varname("__Level", Line),
+    TracesVar = make_varname("__Traces", Line),
+    PidVar = make_varname("__Pid", Line),
+    %% Wrap the call to lager:dispatch_log/6 in case that will avoid doing any work if this message is not elegible for logging
```
macintux: I'm really struggling with this; I may have convinced myself this is necessary previously, but if so, I'm not sure how.

`lager:dispatch_log` and the parse transform do the same (?) sanity checking to duck (if possible) the expensive work embedded in `lager:do_log`.

A function call to the `lager:dispatch_log` shouldn't be significantly more expensive than inlining everything in the parse transform. Is it really worth duplicating everything here?

src/lager_transform.erl
mrallen1: I wasn't sure of the order of transform -> preprocessor or preprocessor -> transform.  It seemed safer to use the naked atom here. But we should give the macro a shot, I suppose.

src/lager_transform.erl
mrallen1: The comment is misleading so we should make it clearer what it happening.  If you read the transform carefully, you will see that the transform **is** `lager:dispatch_log/6` in its AST form. If the message *is* processed further, it's passed on to `lager:do_log/9`.

macintux: I understand, and that's my point: we're inlining `lager:dispatch_log/6` at a non-trivial maintenance cost, and I'm skeptical about the need to do so.

mrallen1: So the proposal is to just inline a call directly to `lager:do_log/9` for everything?  Or inline do_log/9 directly?

macintux: My proposal is to just call `dispatch_log` instead of reimplementing it here. Maybe I'm missing something important.

README.md
macintux: Made some changes. Still need to update tracing docs.

src/lager_transform.erl
mrallen1: OK, had a good long think about this. I thought the whole point of inlining the dispatch logic was to avoid the overhead of needed to copy the parameters to `dispatch_log/6` for every potential message call.  That was, if I recall correctly, one of the areas where lager performance beats out most other things.

Vagabond: Yeah, you want it to be *fast* so unlogged messages are cheap.

macintux: Fair enough. Perils of speculating without benchmarking.

Thanks @Vagabond. I'll continue my double check of the transform code.

src/error_logger_lager_h.erl
macintux: This is no longer relevant: I've special-cased the throttle and no longer break all existing backends.

src/lager_transform.erl
```diff
@@ -75,138 +77,148 @@ walk_clauses(Acc, [{clause, Line, Arguments, Guards, Body}|T]) ->
...
+    %% See lager.erl (lines 89-100) for lager:dispatch_log/6
+    %% case {whereis(Sink), whereis(?DEFAULT_SINK), lager_config:get({Sink, loglevel}, {?LOG_NONE, []})} of
+    {'case',Line,
+         {tuple,Line,
+                [{call,Line,{atom,Line,whereis},[{atom,Line,SinkName}]},
+                 {call,Line,{atom,Line,whereis},[{atom,Line,lager_event}]}, %% <--lager_event is ?DEFAULT_SINK
+                 {call,Line,
+                       {remote,Line,{atom,Line,lager_config},{atom,Line,get}},
+                       [{tuple,Line,[{atom,Line,SinkName},{atom,Line,loglevel}]},
+                        {tuple,Line,[{integer,Line,0},{nil,1}]}]}]},
```
macintux: I assume this should be `{nil,Line}` instead of `{nil,1}`

mrallen1: Good catch. Yes.

mrallen1: Just pushed 6226ef6 to fix that. :cake:

src/lager_transform.erl
```diff
@@ -75,138 +77,148 @@ walk_clauses(Acc, [{clause, Line, Arguments, Guards, Body}|T]) ->
...
+                  [{tuple,Line,
+                          [{var,Line,PidVar},
+                           {var,Line,'_'},
+                           {tuple,Line,[{var,Line,LevelVar},{var,Line,TracesVar}]}]}],
+                  [[{op, Line, 'orelse',
+                    {op, Line, '/=', {op, Line, 'band', {var, Line, LevelVar}, {integer, Line, SeverityAsInt}}, {integer, Line, 0}},
+                    {op, Line, '/=', {var, Line, TracesVar}, {nil, Line}}}]],
+                  [{call,Line,{remote, Line, {atom, Line, lager}, {atom, Line, do_log}},
+                         [{atom,Line,Severity},
+                          Traces,
```
macintux: As far as I can tell this is the right binding, but `Traces` seems confusing as a name for a list of metadata attributes. (I know it predates this PR, but as long as we're revamping things...)

src/lager_transform.erl
mrallen1: Just pushed 060402b and all 199 tests pass successfully. So I guess we're good :)

src/lager_transform.erl
mrallen1: Agree. Added 3223014 to address this. What do you think?

mrallen1:
Closing in favor of #270

PR #265, opened 2015-05-17
Add abbreviated severity 'sev' available to formatter

saleyn:


mrallen1:
Could you help me understand the intention of this change?

saleyn:
The intention is to be able to have an option of more succinct aligned log levels in the form:
```
{lager_default_formatter,
                        [time, " [",sev,"] ", message, " [", module, ":", line, "]\n"]}]}
```
That would produce:
```
10:12:00 [W] Test warnings [some_module:123]
```

Vagabond:
This looks fine to me. If all the tests pass and look good I don't have a problem with this landing.

saleyn:
All the tests pass except for two, which don't pass even without applying my patch, so I believe it's benign.

mrallen1:
Just committed d6cb3bf on intg and all the tests pass. Thanks.

PR #266, opened 2015-05-19
Cherry pick OTP 17 work

mrallen1:
This branch cherry-picks OTP 17 patches from `feature/otp-17` into a clean branch along with selected bug fixes so we create a stable 2.x branch of lager that retains all current semantics and supports OTP 17 and fixes all known test failures.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -1,9 +1,33 @@
-{erl_opts, [debug_info, warn_untyped_record]}.
+%% -*- erlang -*-
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2011-2015 Basho Technologies, Inc.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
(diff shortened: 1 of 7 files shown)


rebar.config
```diff
@@ -1,9 +1,33 @@
...
+
+{erl_opts, [
+    debug_info,
+    warn_untyped_record
+]}.
 {deps, [
-        {goldrush, "0\.1\.6",
-            {git, "git://github.com/DeadZen/goldrush.git", {tag, "0.1.6"}}}
-       ]}.
+    {goldrush, ".*", {git, "git://github.com/DeadZen/goldrush.git", {branch, "master"}}}
```
mrallen1: @tburghart I don't understand why we want to unpin this. Seems to compile fine on OTP-17 for me. Was there another reason for this change?

tburghart: One of the first things I did in all repos in which I created OTP-17 branches was to use develop/master (as appropriate) for their dependencies, since lots of them included relevant OTP-17 patches. I didn't evaluate each separately unless I ran into an issue, and took the approach that we might subsequently rein them in if we needed to, but I wanted to get the current work in them.

Riak EE uses locked-deps for releases, so unless there's a reason *not* to use the current work,  I'd rather pick up patches as the maintainers apply them. That said, I don't assume they'll all stay "unpinned" - this is a development branch, after all.

Vagabond: I don't recommend we unpin goldrush, I know @DeadZen has recently been making changes, and we don't want to break tags (or remove the ability to do reproducible builds) because of a floating dependancy.

mrallen1:
Going to reopen this against the 2.x branch and re-pin the deps to the newest version of goldrush.

PR #267, opened 2015-05-23
fix lager:stop_trace

pvmart:
Fix stop_trace_int so it can remove an event handler


mrallen1:
Thanks for the fix! Sorry it has taken us so long to respond to it. I have a PR open to solve the problem as the code currently stands.

mrallen1:
Superseded by #308 and #306

Issue #274, opened 2015-06-23
Lager crash log path does not honor log_root [JIRA: RIAK-1882]


Vagabond:

mrallen1:
This is for the release version?  I think we could get a fix in place for this before all of the review-mergy-mergy stuff that's still not finished.

Vagabond:
Yeah, I just ran into it, affects latest 2.x and probably upcoming 3.x

mrallen1:
Just added some crash_log unit tests with log_root on 3.x and they all pass. :cake: Going to check on 2.x now.

mrallen1:
OK, so these unit tests are not capturing the failure.  Can you be more specific about what's failing?

mrallen1:
@Vagabond if you have a moment to update this ticket - I'd really like to get this resolved for you.  My test case is not capturing a failure, so if you can provide additional information it would be really helpful. Thanks.

Vagabond:
Yeah, I'm kind of baffled, reading the code, why I had this not working. As far as your testcases and the code look, this should work. Might just have been misdiagnosis or a problem with lager getting reverted to an older version on my part.

Vagabond:
I think we can close this.

mrallen1:
OK. Thanks for the clarification. I'll get #277 rebased to master for a 3.next release soon.

PR #277, opened 2015-06-24
Add unit tests for log_root in the crash_log module

mrallen1:
Trying to address #274 (RIAK-1882), but these tests do not capture a failure.


mrallen1:
This needs to be rebased on to master. Closing for now.

Issue #273, opened 2015-06-22
Lager performance improvements [JIRA: RIAK-1883]


robertkowalski:
Hi and first of all thanks for lager!

I observed that lager is one of the main bottleneck in our database http handlers (see [1]). The database operations at the end of the x axis seem to take a minority of the time it take us to log (see at the end of the flamegraph which is also provided as svg in [2])

It seems that if we try to let lager format our log message, it takes a huge amount of time (see [2]) - but if we pass in a simple list, it gets really fast again (~4% speed improvement)

Are there any ways to speed that up from the lager side which does not require us to handcraft the log message using a list?

1. Maybe we can move the format call for log messages [3] into the gen_event_server [4] and by doing that decouple it from the client call?

2. Maybe we can work on the formatter making it more performant? 

What do you think?

[1] https://github.com/apache/couchdb-couch/pull/57
[2] http://robert-kowalski.de/assets/data/2015-06-01-erlang-perf-2/flame--second--unpatched.svg
[3] https://github.com/basho/lager/blob/95f5a133af6b5b42635f7079b626ac1886eccde2/src/lager.erl#L109
[4] https://github.com/basho/lager/blob/95f5a133af6b5b42635f7079b626ac1886eccde2/src/lager.erl#L117

mrallen1:
Seems like the root of the perceived slowness is in lager_trunc_io - which is not necessarily built for performance but for safety.  One of the reasons it exists is so can truncate long (>4096 bytes) log messages which get sent by OTP stuff (gen_event or gen_fsm for example) which dumps its ENTIRE state when it fails. That might be upwards of 10MB or more depending on what its processing. As @Vagabond explains in his [tech talk about lager](https://www.youtube.com/watch?v=8BNpOHFvg_Q), those types of error messages are ticking time bombs for your application availability.

The forthcoming [lager 3.0 branch](https://github.com/basho/lager/pull/270) we are preparing to release does add some performance benefits, most clearly seen when there is more than one sink configured for messages to drain.   I know it may not *directly* address your performance concerns about message formatting, but our [benchmarks](https://gist.github.com/mrallen1/681b2b0f0e55e56e8e29) do show a modest improvement in logging performance in the 3.x branch even when there's only 1 sink configured.

All of that not withstanding, I am definitely interested in wringing more performance out of lager. So I am all ears for ideas (or PRs) you have for the formatting code which retains its safety characteristics and is also speedier.

cc @macintux

mrallen1:
[...]

PR #278, opened 2015-06-25
Add an unsafe_formatting code path

mrallen1:
This branch adds a new `_unsafe` path for log messages with a reasonable bounded size when its not necessary for lager to ensure that log messages don't kill an application because they're so huge.

### DIFF
--- a/src/lager.erl
+++ b/src/lager.erl
@@ -26,12 +26,13 @@
 %% API
 -export([start/0,
         log/3, log/4, log/5,
+        log_unsafe/4,
         md/0, md/1,
         trace/2, trace/3, trace_file/2, trace_file/3, trace_file/4, trace_console/1, trace_console/2,
         name_all_sinks/0, clear_all_traces/0, stop_trace/1, stop_trace/3, status/0,
         get_loglevel/1, get_loglevel/2, set_loglevel/2, set_loglevel/3, set_loglevel/4, get_loglevels/1,
         update_loglevel_config/1, posix_error/1, set_loghwm/2, set_loghwm/3, set_loghwm/4,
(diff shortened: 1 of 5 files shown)


mrallen1:
Inspired by #273

mrallen1:
@JeetKunDoug @jvoegele After our discussion last week would you please review this PR so it can be merged into 3.x?

mrallen1:
cc @macintux

src/lager.erl
```diff
@@ -486,6 +504,23 @@ safe_format(Fmt, Args, Limit, Options) ->
...
     safe_format(Fmt, Args, Limit, [{chomp, true}]).
 
+%% @private Print the format string `Fmt' with `Args' without a size limit.
+%% This is unsafe because the output of this function is unbounded.
+%%
+%% Log messages with unbounded size will kill your application dead as
+%% OTP mechanisms stuggle to cope with them.  So this function is
+%% intended <b>only</b> for messages which have a reasonable bounded
+%% size before they're formatted.
+%%
```
JeetKunDoug: I wonder if this is really a strong enough warning, and if it shouldn't have a README change as well - would be more likely to be seen there. Is there a set of README changes planned to go with this?

JeetKunDoug:
Other than the above note on warning/guidance for unsafe stuff (which I'd be happy to help write/review), this looks fine to me.

jvoegele:
I agree with @JeetKunDoug's comment above, but other than that :+1:

mrallen1:
Yeah, thanks for the reminder. I'll add a blurb to the README as a new commit here in a second.

mrallen1:
OK. I will add some language to the README in a second as a new commit.

mrallen1:
OK, pushed some text about `_unsafe`

JeetKunDoug:
:+1: 2e81dc3

mrallen1:
@borshop merge

PR #279, opened 2015-06-29
Use encrypted, authenticated, HTTPS over Git for fetching goldrush.

ahf:
This patch shouldn't impact people's CI system as both git:// and
https:// should work nicely without authentication.


umbec:
:+1: I just forked to make the same patch...

macintux:
We've started discussing this internally. For now, you may find this git configuration option handy:

http://erlang.org/pipermail/erlang-questions/2014-March/078402.html

mrallen1:
Master already does this, so closing this PR.

PR #281, opened 2015-07-24
Extend lager_msg with format and args

zdaniel86:
Hello guys. I want to add extra fields (format and args) to lager_msg to be able to evaluate formatting in lager backend module.

It can be useful when backend module want to process it differently against io_lib:format


Vagabond:
This is pretty dangerous, as you potentially end up copying a 5mb string or something if some of your format arguments are not what you expect.

macintux:
I'm inclined to agree that @Vagabond's concern makes this a dangerous idea, and one I'd rather not incorporate without more precautions than I believe are available to us inside lager itself, since the 3rd party backend would have the reins.

Closing.

Issue #280, opened 2015-07-24
function_clause error msgs on 3.0.0 version. [JIRA: RIAK-2042]


snaiper80:
I tried to upgrade the lager's version in my projects to tag 3.0.0

But I got a some crash log on console as below.
How can I fix this error? No issues on lager 2.X.

```
Erlang/OTP 17 [erts-6.4.1] [source-381fb6c] [64-bit] [smp:2:2] [async-threads:6] [hipe] [kernel-poll:true]


=ERROR REPORT==== 24-Jul-2015::11:51:28 ===
** gen_event handler log_mf_h crashed.
** Was installed in error_logger
** Last event was: get_loglevel
** When handler state == {state,"log/sasl",10485760,5,25753,5,
                                {file_descriptor,prim_file,{#Port<0.5227>,18}},
                                [],#Fun<sasl.0.89613933>}
** Reason == {function_clause,
                 [{log_mf_h,handle_call,
                      [get_loglevel,
                       {state,"log/sasl",10485760,5,25753,5,
                           {file_descriptor,prim_file,{#Port<0.5227>,18}},
                           [],#Fun<sasl.0.89613933>}],
                      [{file,"log_mf_h.erl"},{line,164}]},
                  {gen_event,server_call_update,3,
                      [{file,"gen_event.erl"},{line,638}]},
                  {gen_event,server_call,4,
                      [{file,"gen_event.erl"},{line,587}]},
                  {gen_event,handle_msg,5,[{file,"gen_event.erl"},{line,258}]},
                  {proc_lib,init_p_do_apply,3,
                      [{file,"proc_lib.erl"},{line,237}]}]}

=ERROR REPORT==== 24-Jul-2015::11:51:28 ===
** gen_event handler sasl_report_file_h crashed.
** Was installed in error_logger
** Last event was: get_loglevel
** When handler state == {<0.1297.0>,"log/sasl-error.log",error}
** Reason == {error,bad_query}
Eshell V6.4.1  (abort with ^G)
(bird@127.0.0.1)1>
```

my sys.config file is...
```
[
 %% SASL config
 {sasl, [
         {sasl_error_logger, {file, "log/sasl-error.log"}},
         {errlog_type, error},
         {error_logger_mf_dir, "log/sasl"},      % Log directory
         {error_logger_mf_maxbytes, 10485760},   % 10 MB max file size
         {error_logger_mf_maxfiles, 5}           % 5 files max
        ]},

 %% Lager config
 {lager, [

[...]

PR #282, opened 2015-07-29
Add test case and fix for #280

mrallen1:
This PR contains a test case (which **must** be run as the first test case because reasons :cry:) and a fix for the crash report in ticket #280 (RIAK-2042).

The root cause is when lager 3.0 takes OTP sink names like `error_logger` and tries to interrogate them for their current loglevel. Of course OTP sinks do not implement log levels, so this causes a function clause crash.

We fixed this issue by adding a special clause testing for `error_logger` and simply returning `true` which is the expected value for this function to return (as it updates an internal ETS table for other sinks and ETS returns `true` when it completes successfully.)


javajolt:
:+1: 9c0e141

mrallen1:
@borshop merge

PR #292, opened 2015-09-02
Accept 255 (or 0xFF in hexadecimal) for format strings

shino:
At least for ISO 8859-1 (aka Latin-1) [1], 255 is one of its
characters.

Behavior before this commit:

```
9> lager:log(info, self(), ["254 is ok: ", 254]).
ok
17:55:11.624 [info] 254 is ok: þÿ
10> lager:log(info, self(), ["255 is not considered as \"character\" (sadpanda)", 255]).
ok
17:55:15.080 [info] FORMAT ERROR: ["255 is not considered as \"character\" (sadpanda)",255] []
```

[1] https://en.wikipedia.org/wiki/ISO/IEC_8859-1

### DIFF
--- a/src/lager_stdlib.erl
+++ b/src/lager_stdlib.erl
@@ -39,7 +39,7 @@ string_p([]) ->
 string_p(Term) ->
     string_p1(Term).
 
-string_p1([H|T]) when is_integer(H), H >= $\s, H < 255 ->
+string_p1([H|T]) when is_integer(H), H >= $\s, H < 256 ->
     string_p1(T);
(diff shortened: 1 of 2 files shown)


voluntas:
+1

mrallen1:
We need to forward port this to 3.x too.

shino:
Is it OK to merge? I guess +1 from Mr. V is sufficient :smile:  but wait for another one from internal basho.

mrallen1:
I didn't realize it hadn't been merged.

shino:
Thanks @mrallen1 !

PR #294, opened 2015-10-16
Add option to print all metadata in the default formatter

Vagabond:
Metadata can be printed with standard or custom internal seperators ad field
seperators. The defaults are "=" and " ":

  "key1=value1 key2=value2"


mrallen1:
LGTM. Thanks!

PR #302, opened 2015-12-06
Tracing: support identified backends

g-andrade:
Tracing: support identified backends

Allow for identified backends (`{module(), term()}`) on `lager:trace/2` and `lager:trace/3`, using existing support for 'lager_file_backend' as a basis.

Use case: having multiple instances of a custom backend, each with a different filter, added during runtime.

### DIFF
--- a/src/lager.erl
+++ b/src/lager.erl
@@ -198,49 +198,13 @@ validate_trace_filters(Filters, Level, Backend) ->
     }.
 
 trace_file(File, Filter) ->
-    trace_file(File, Filter, debug, []).
+    trace({lager_file_backend, File}, Filter).
 
-trace_file(File, Filter, Level) when is_atom(Level) ->
-    trace_file(File, Filter, Level, []);
-
-trace_file(File, Filter, Options) when is_list(Options) ->
-    trace_file(File, Filter, debug, Options).
+trace_file(File, Filter, LevelOrOptions) ->
+    trace({lager_file_backend, File}, Filter, LevelOrOptions).
 
 trace_file(File, Filter, Level, Options) ->
-    FileName = lager_util:expand_path(File),
-    case validate_trace_filters(Filter, Level, {lager_file_backend, FileName}) of
-        {Sink, {ok, Trace}} ->
-            Handlers = lager_config:global_get(handlers, []),
-            %% check if this file backend is already installed
-            Res = case lists:keyfind({lager_file_backend, FileName}, 1, Handlers) of
-                      false ->
-                          %% install the handler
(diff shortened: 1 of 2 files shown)


mrallen1:
Thanks for this PR. As a quick browse, it looks great. We are super busy with a product release right now, so I would ask for some extra patience as other Basho engineers need to look this over.

mrallen1:
We triaged a bunch of things today but ran out of time to really get our heads around this change today. I'm sorry. It's going to have to wait until after the holidays.

g-andrade:
OK, no problem. Thank you in anycase! 2016 it is.


Sent from my phone. Please excuse my brevity.

g-andrade:
Any news?

mrallen1:
Hi - we have more triage/PR review scheduled first thing tomorrow. (Today was a US holiday.)

mrallen1:
Just wanted to let you know, we are planning to merge this but it we are going to have to modify it to accommodate the bug fixes in #306 with respect to stopping traces correctly and updating application state correctly.  I am planning to work on that tomorrow. We should have a final review/merge Friday.

mrallen1:
@macintux Thought we'd hit a merge conflict here but we didn't - everything looks clean and passes all tests and dialyzer is green too.

g-andrade:
Great!

macintux:
I am a bit concerned with the interactions with multiple sink support.

The file backend is tied to a single sink, and the new `trace_identified_backend` extends that limitation to any backend using the new mechanism.

Also thinking about the way we use the global `handlers` configuration to track files, and wondering whether extending that to other modules has any undesirable impact.

mrallen1:
I did review this again today. We need to make sure that this generalization doesn't have any unintended consequences before merging it and that's going to require investigation I don't have time to do right now.

I think @macintux is right that we ought to be wary of specifically backing ourselves into the limitation of a single sink since that's kind of a flaw in the current multi-sink support.

g-andrade:
I had forgotten about this entirely... :-) 
I'm no longer in need of the mentioned functionality, nor am I still familiar with the actual solution (I would have to review what I did at the time.)
Should I leave this PR as is, for someone else that might want to take it, or should I close it instead?

mrallen1:
@g-andrade Thanks for the follow up. If you don't mind, let's leave it open for a while (say, 6 months or so) and if no one has time to deal with it, I'll close it.  I think the generalization idea is a good one but I want to be sure there's aren't any unexpected consequences.

mrallen1:
OK, going to close this for now.

PR #303, opened 2015-12-09
Support for separate sink for error_logger messages

hairyhum:
If `error_logger_lager_event` sink is defined, all `error_logger` messages goes to this sink, otherwise - default sink. 
This provides opportunity to separate application logs and `error_logger` logs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,23 @@ application variable. If set to `undefined` it is not written at all.
 Messages in the crash log are subject to a maximum message size which can be
 specified via the `crash_log_msg_size` application variable.
 
+Messages from `error_logger` will be redirected to `error_logger_lager_event` sink
+if it is defined so it can be redirected to another log file.
+For example:
+
+```
+[{lager, [
+         {extra_sinks,
+          [
+           {error_logger_lager_event, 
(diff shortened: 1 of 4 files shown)


mrallen1:
I like this idea. Thanks for the PR.

Would you please add some tests into [test/lager_backend_test.erl](https://github.com/basho/lager/blob/master/test/lager_test_backend.erl) to validate that it functions as expected?

Also we are very busy with a product release right now, so this PR may take some extra time to review as most of my colleagues are tied up with that right now.

hairyhum:
Added tests. 
And there are failing tests with warning_map() I didn't figure out, but it failing in upstream as well.

mrallen1:
Thanks for adding the tests. We'll take a look at `warning_map()` when we get a chance.

mrallen1:
`warning_map()` is broken because OTP changed the output from versions previous to 18.  It used to return `error` and now it returns `warning` and lager still expects `error`. That's why these tests fail.

mrallen1:
:+1: 2248a2a

mrallen1:
@borshop merge

PR #305, opened 2015-12-14
Option to not reformat error_logger messages

hairyhum:
Application variable `error_logger_format_raw` when set to `true` prevents error_logger messages from reformatting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ Lager is also supplied with a `error_logger` handler module that translates
 traditional erlang error messages into a friendlier format and sends them into
 lager itself to be treated like a regular lager log call. To disable this, set
 the lager application variable `error_logger_redirect` to `false`.
-You can also disable reformating for OTP and Cowboy messages by setting variable
+You can also disable reformatting for OTP and Cowboy messages by setting variable
 `error_logger_format_raw` to `true`.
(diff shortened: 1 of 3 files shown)


mrallen1:
Thanks for the idea. 

After some consideration, we'd prefer to see the "raw" option in the module state rather than something that's looked up inline during the `log_event/2` call.  You can pattern match on the value in the function head and the application won't have to constantly lookup the setting while lager is processing log messages.

README.md
```diff
@@ -222,6 +222,8 @@ Lager is also supplied with a `error_logger` handler module that translates
 traditional erlang error messages into a friendlier format and sends them into
 lager itself to be treated like a regular lager log call. To disable this, set
 the lager application variable `error_logger_redirect` to `false`.
+You can also disable reformating for OTP and Cowboy messages by setting variable
+`error_logger_format_raw` to `true`.
```
macintux: Typo: "reformatting".

Other than that, this looks fine. I'm not thrilled that `error_logger_lager_h:log_event` just got hairier, but it's not significantly worse than it already was. Problem for another day.

Will give it final approval with the typo fix.

mrallen1:
@borshop merge

PR #306, opened 2015-12-17
Fix internal tuple argument order

mrallen1:
As identified in PR #267, the argument ordering in the internal representation of a trace tuple as returned by `validate_trace_filters` is different from the external representation, and we had the external ordering in the function head.

This prevented backends from being stopped or removed if no other traces were active for that backend.


mrallen1:
@borshop merge

Issue #296, opened 2015-11-09
Configure tracing for custom sink [JIRA: RIAK-2276]


nkmprog:
Not sure if this is an issue but here goes nothing. I'm trying to configure [tracing](https://github.com/basho/lager#tracing) for a custom sink in [lager 3.0.1.](https://github.com/basho/lager/tree/3.0.1)

I've been attemting to so in app.config - knowingly avoiding the use of the runtime configuration by lager:trace_file/3. Using [this issue](https://github.com/basho/lager/issues/207) as some sort of reference I have the following app.config.

    {lager, [
            {handlers,
             [{lager_console_backend, info},
              {lager_file_backend,
                ...
              }
      
             ]},
            {extra_sinks,
              [{custom_lager_event, [
                     {handlers, [{lager_console_backend, info},
                                 {lager_file_backend,
                                  [{file, "log/custom.log"},
                                   {level, info},
                                   {formatter, lager_default_formatter},
                                   {formatter_config, [message, "\n"]}
                                 ]}]}
                  ]}]
            },
            {traces, [{ {lager_file_backend, "log/login.log"},
                        [{sink, custom_lager_event}, {type, tag}],
                        info }]
            }
           ]}

However upon application start i get the following error

    =INFO REPORT=
    application: lager
    exited: {bad_return,
                {{lager_app,start,[normal,[]]},
                 {'EXIT',
                     {{badmatch,{error,noproc}},
                      [{lager_app,start_handler,3,
                           [{file,"src/lager_app.erl"},{line,96}]},
                       {lager,trace_file,4,
                           [{file,"src/lager.erl"},{line,225}]},
                       {lager_app,'-add_configured_traces/0-fun-0-',1,
                           [{file,"src/lager_app.erl"},{line,254}]},
                       {lists,foreach,2,[{file,"lists.erl"},{line,1336}]},
                       {lager_app,add_configured_traces,0,
                           [{file,"src/lager_app.erl"},{line,253}]},
                       {lager_app,start,2,
                           [{file,"src/lager_app.erl"},{line,204}]}

I should specify that the sink is properly configured and running without the trace config. Configuring a trace to the default and custom sinks at runtime also works.

[...]

PR #307, opened 2015-12-17
Start traces after extra sinks are started

mrallen1:
Previously, if you configured a trace to use an extra sink, you'd get a noproc error. Addresses issue #296 on github.


mrallen1:
@borshop merge

PR #311, opened 2016-01-11
API to explicitly call log rotation

hairyhum:
Can be used with `logrotate` scripts to have more control over log rotation, while still using lager log rotation features.

### DIFF
--- a/src/lager.erl
+++ b/src/lager.erl
@@ -28,6 +28,7 @@
         log/3, log/4, log/5,
         log_unsafe/4,
         md/0, md/1,
+        rotate_file/1, rotate_sink/1, rotate_all/0,
         trace/2, trace/3, trace_file/2, trace_file/3, trace_file/4, trace_console/1, trace_console/2,
         list_all_sinks/0, clear_all_traces/0, stop_trace/1, stop_trace/3, status/0,
         get_loglevel/1, get_loglevel/2, set_loglevel/2, set_loglevel/3, set_loglevel/4, get_loglevels/1,
@@ -594,3 +595,26 @@ pr_stacktrace(Stacktrace) ->
 pr_stacktrace(Stacktrace, {Class, Reason}) ->
     lists:flatten(
         pr_stacktrace(Stacktrace) ++ "\n" ++ io_lib:format("~s:~p", [Class, Reason])).
+
+rotate_file(FileName) ->
+    Handlers = lager_config:global_get(handlers),
+    RotateHandlers = lists:filter(
+        fun ({{lager_file_backend, FN}, _, _}) ->
+                FN == FileName orelse 
+                lager_util:expand_path(FN) == lager_util:expand_path(FileName);
+            (_) -> false
+        end,
+        Handlers),
+    lager_file_backend:rotate_handlers(RotateHandlers).
+
+rotate_sink(Sink) ->
+    Handlers = lager_config:global_get(handlers),
+    RotateHandlers = lists:filter(
+        fun({_,_,S}) -> S == Sink; 
+           (_) -> false 
+        end, 
+        Handlers),
+    lager_file_backend:rotate_handlers(RotateHandlers).
+
+rotate_all() -> 
+    lager_file_backend:rotate_handlers(lager_config:global_get(handlers)).
--- a/src/lager.erl
+++ b/src/lager.erl
@@ -605,7 +605,7 @@ rotate_file(FileName) ->
             (_) -> false
         end,
         Handlers),
-    lager_file_backend:rotate_handlers(RotateHandlers).
+    rotate_handlers(RotateHandlers).
 
 rotate_sink(Sink) ->
     Handlers = lager_config:global_get(handlers),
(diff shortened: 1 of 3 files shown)


mrallen1:
Nice idea. 

This feature is going to need at least a couple of tests and some documentation before we can look at merging it.

Thank you.

hairyhum:
Also I wanted to ask why during log rotation (with `{rotate, File}` message) file handle is not being closed? Is it on purpose?

mrallen1:
There's a PR - #312 to fix that. I'm not sure. Probably an oversight - or maybe it was implicitly getting closed when the file handle changed.

dumbbell:
Hi!

Maybe a bit of context here can help you during the review of this patch. @hairyhum worked on changing RabbitMQ to use Lager as the logging subsystem, replacing a homebrew module. RabbitMQ comes with a CLI to force a log rotation among other things. We would like to keep this feature, that's why a public API in Lager would be welcome.

Our use would be something like:
```erlang
% For each sink:
lager:log(Sink, info, self(), "About to rotate", []),
lager:rotate(Sink),
lager:log(Sink, info, self(), "New file created", [])
```

So far we played with this by sending `{rotate_file, File}` to sinks having a `lager_file_backend` in their handlers. This and the patch proposed in #312 achieve what we woud like. #312 is necessary so that the message "About to rotate" ends up in the old file and "New file created" is written to the new file. Without #312, both messages go to the old file (as well as some subsequent messages); the new file is created only after the next `check_interval`.

@hairyhum: About your patch, instead of explicitely limit log rotation to `lager_file_backend` in `lager.erl`, I would simply send an event or call all handlers without filtering. Thus, any handler can implement the feature if it makes sense. Also, in our specific case, I would prefer the call to be synchronous so we know when the rotation did happen; I don't know what you would prefer, @mrallen1.

hairyhum:
@dumbbell it's good idea to let all handlers do rotation. Here is implementation using `gen_event:call` to allow handlers to define rotation. It's better than just send message.

mrallen1:
I'll be looking into this PR in depth tomorrow, with a final pass with my colleagues on Friday.

test/lager_rotate.erl
```diff
@@ -0,0 +1,135 @@
...
+
+                have_no_log(SinkFile, <<"Sink test message 1">>),
+                have_log(SinkFile, <<"Sink test message 2">>),
+
+                have_log(File1Old, <<"Test message 1">>),
+                have_no_log(File1Old, <<"Test message 2">>),
+
+                have_log(File2Old, <<"Test message 1">>),
+                have_no_log(File2Old, <<"Test message 2">>)
+
```
macintux: Should check for contents of old sink file here too:

```

              have_log(SinkFileOld, <<"Sink test message 1">>),
              have_no_log(SinkFileOld, <<"Sink test message 2">>),
```

src/lager.erl
```diff
@@ -594,3 +595,34 @@ pr_stacktrace(Stacktrace) ->
...
+
+rotate_all() -> 
+    rotate_handlers(lager_config:global_get(handlers)).
+
+
+rotate_handlers(Handlers) ->
+    [ rotate_handler(Handler) || Handler <- Handlers ].
+
+rotate_handler({Backend, Handler, Sink}) ->
+    gen_event:call(Sink, Backend, rotate, infinity).
```
macintux: Looks like any 3rd party backends will not know about the new `rotate` message and cause this to lock up indefinitely.

Perhaps it makes more sense to use a `gen_event:cast` for this.

src/lager.erl
```diff
@@ -594,3 +595,34 @@ pr_stacktrace(Stacktrace) ->
 pr_stacktrace(Stacktrace, {Class, Reason}) ->
     lists:flatten(
         pr_stacktrace(Stacktrace) ++ "\n" ++ io_lib:format("~s:~p", [Class, Reason])).
+
+rotate_file(FileName) ->
+    Handlers = lager_config:global_get(handlers),
+    RotateHandlers = lists:filter(
+        fun ({{lager_file_backend, FN}, _, _}) ->
```
macintux: Worth considering whether this should be an underscore match against the module name so that other file-wrapping handlers could be used. If so, `lager_util:expand_path` would need to gracefully handle non-string arguments, or those would have to be wrapped in catch.

Alternatively or additionally, exposing `rotate_handler` (perhaps with a parameter list that makes it look more like `rotate_file`) would allow other backends to participate in this (without the user having to invoke `rotate_sink` or `rotate_all` to sidestep our constraints).

src/lager.erl
hairyhum: It's common practice that `gen_event` modules have empty `handle_call` to skip unknown messages and don't block them.
`gen_event:call` is used here on purpose to wait until the file will be actually rotated, considering #316

src/lager.erl
hairyhum: There is no standard file-wrapping handler spec. So It's unsafe to assume any handler with 2-element tuple to be a file handler.
By exposing `rotate_handler`, do you mean something like this:
```
rotate_handler(Backend) ->
    rotate_handler(Sink, Backend).
rotate_handler(Sink, Backend) ->
    gen_event:call(Sink, Backend, rotate, ?TIMEOUT).
```

src/lager.erl
macintux: I see, wasn't aware of the common practice, thanks.

hairyhum:
Added `rotate_handler/1,2` to rotate any handlers. File backend is rotated using `rotate_handler({lager_file_backend, "file.log"})`

macintux:
I really appreciate the work you've put into this. Nice job.

mrallen1:
@borshop merge

PR #313, opened 2016-01-19
Bump goldrush to 0.1.8 for not equal filters

mrallen1:
The upstream goldrush library added support for not equal filters. All lager tests pass.


mrallen1:
@borshop merge

PR #312, opened 2016-01-11
lager_file_backend: Close file after time-based rotation

dumbbell:
Before, the old file was still being used until the next `check_interval`, where the file was recreated and reopened.

PR #316, opened 2016-01-20
Close file right after rotation

mrallen1:
This PR supersedes #312 and includes a trivial fix for an unmatched function return that makes dialyzer cranky.


bsparrow435:
:+1: 063d628549b9bd1ec934f3be4ae715e8a17e6bb4

mrallen1:
@borshop merge

PR #321, opened 2016-02-21
Support typed records newly exposed in OTP 19

weisslj:
Otherwise `lager_transform` fails after https://github.com/erlang/otp/commit/de9012628a6b0e97d2f1325bf2f72817f69f84ee

The error message is:

    test/pr_nested_record_test.erl: error in parse transform 'lager_transform': {function_clause,
                                                 [{lager_transform,
                                                   '-walk_ast/2-fun-0-',
                                                   [{typed_record_field,
                                                     {record_field,5,
                                                      {atom,5,field1}},
                                                     {type,5,term,[]}}],
                                                   [{file,
                                                     "src/lager_transform.erl"},
                                                    {line,62}]},

### DIFF
--- a/src/lager_transform.erl
+++ b/src/lager_transform.erl
@@ -59,15 +59,18 @@ walk_ast(Acc, [{function, Line, Name, Arity, Clauses}|T]) ->
     walk_ast([{function, Line, Name, Arity,
                 walk_clauses([], Clauses)}|Acc], T);
 walk_ast(Acc, [{attribute, _, record, {Name, Fields}}=H|T]) ->
-    FieldNames = lists:map(fun({record_field, _, {atom, _, FieldName}}) ->
-                FieldName;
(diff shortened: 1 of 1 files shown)


mrallen1:
Thanks for the patch!

okeuday:
Also ran into this, this patch appears to be required for any lager use with 19.0.

weisslj:
I did not mention it, but this patch can be applied already, it works with OTP 18 as well.

okeuday:
I haven't seen problems using the patch with OTP R16 and 17.x either.  Using it with `2.0.3`.

macintux:
I appreciate the feedback, helpful to know it hasn't caused problems with earlier releases. I'll take a closer look very soon.

macintux:
:+1: d35670e

macintux:
@borshop merge

PR #325, opened 2016-03-09
fix function calls not available in R15

zzantozz:
In the rest of lager, application:get_env/3 isn't used. `lager_app` specifically calls
out that the call isn't available in R15. This module, `error_logger_lager_h`, is the
only module that uses it, and this is a relatively simple fix.

### DIFF
--- a/src/error_logger_lager_h.erl
+++ b/src/error_logger_lager_h.erl
@@ -73,7 +73,10 @@ set_high_water(N) ->
 -spec init(any()) -> {ok, #state{}}.
 init([HighWaterMark, GlStrategy]) ->
     Shaper = #lager_shaper{hwm=HighWaterMark},
-    Raw = application:get_env(lager, error_logger_format_raw, false),
+    Raw = case application:get_env(lager, error_logger_format_raw) of
+        {ok, Value} -> Value;
+        undefined -> false
+    end,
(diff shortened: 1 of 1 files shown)


Vagabond:
I don't understand what this is trying to fix?

Vagabond:
Oh, you left it as /3 in one place, line 110.

I'm not sure R15 support is terribly interesting to support at his point.

zzantozz:
Oops, [fixed that](https://github.com/zzantozz/lager/commit/53d8f3914b67547570320042fa531dc32c3653ba). Thanks. Yeah, I know it's old. It's what my project is currently working to get off of. We'll be up to date soon, but I figured this is a small change, and lager 3 seems to work great on R15 otherwise.

Vagabond:
I have no objection, as long as it is clear that R15 is not officially supported.

zzantozz:
So.... what needs to happen to get this merged or closed? I'd like to get it resolved so I'll know if I need to maintain a temporary fork of this until we get upgraded off of R15.

mrallen1:
We'll be doing another round of triage and PR code review soon. Thanks for the PR.

tonyrog:
R15 is important since it is still default on some platforms. Raspberry pi is one.
Please fix.

Issue #324, opened 2016-03-09
lager_event message queue growth  [JIRA: RIAK-2433]


blt:
Hi folks!

We've hit instances where a burst of traffic into lager will cause the lager_event message queue to grow substantially, to the point where lager is still emptying out several hours later. This causes a system-wide slowdown as lager has switched synchronous mode to cope. 

A bit goofy, but something that's bit us a few times as our system has a bursty traffic pattern. 

I've put together a repository to reproduce, called [burst](https://github.com/blt/burst). You can see it in action here:

```
(burst@walden)1> spawn(burst, run, [10000,100]), recon:proc_window(message_queue_len, 10, timer:seconds(10)).
[{<0.311.0>,28728,
  [lager_event,
   {current_function,{lager_default_formatter,make_printable,
                                              1}},
   {initial_call,{proc_lib,init_p,5}}]},
 {<0.1040.0>,0,
  [{current_function,{gen,do_call,4}},
   {initial_call,{burst,wrkr,3}}]},
 {<0.1630.0>,0,
  [{current_function,{gen,do_call,4}},
   {initial_call,{burst,wrkr,3}}]},
 {<0.2220.0>,0,
  [{current_function,{gen,do_call,4}},
   {initial_call,{burst,wrkr,3}}]},
 {<0.2810.0>,0,
  [{current_function,{gen,do_call,4}},
   {initial_call,{burst,wrkr,3}}]},
 {<0.7981.0>,0,
  [{current_function,{gen,do_call,4}},
   {initial_call,{burst,wrkr,3}}]},
 {<0.8571.0>,0,
  [{current_function,{gen,do_call,4}},
   {initial_call,{burst,wrkr,3}}]},
 {<0.9161.0>,0,
  [{current_function,{gen,do_call,4}},
   {initial_call,{burst,wrkr,3}}]},
 {<0.3852.0>,0,
  [{current_function,{gen,do_call,4}},
   {initial_call,{burst,wrkr,3}}]},
 {<0.4442.0>,0,
  [{current_function,{gen,do_call,4}},
   {initial_call,{burst,wrkr,3}}]}]
```

The challenge we have is that there's not much control on the user side once `lager_event` backs up. For our use case, once you blow past some higher high-water-mark, it'd be desirable just to start dropping log messages, rather than pool them up for later emission. 

We'd be happy to do the work. Would this be in-line with lager's handling philosophy? 

/cc @JoseManuelRodriguez @miriampena

[...]

Issue #275, opened 2015-06-24
18.0 test failures [JIRA: RIAK-1886]


macintux:
Via @mrallen1 

```
 lager_test_backend: error_logger_redirect_test_ (info messages with unicode characters in Args are printed)...ok
  lager_test_backend: error_logger_redirect_test_ (warning messages with unicode characters in Args are printed)...*failed*
in function lager_test_backend:'-error_logger_redirect_test_/0-fun-60-'/2 (test/lager_test_backend.erl, line 1000)
in call from lager_test_backend:'-error_logger_redirect_test_/0-fun-63-'/0 (test/lager_test_backend.erl, line 1000)
**error:{assertEqual,[{module,lager_test_backend},
              {line,1000},
              {expression,"Level"},
              {expected,16},
              {value,8}]}
  output:<<"">>

  lager_test_backend: error_logger_redirect_test_ (warning messages are printed at the correct level)...*failed*
in function lager_test_backend:'-error_logger_redirect_test_/0-fun-64-'/2 (test/lager_test_backend.erl, line 1012)
in call from lager_test_backend:'-error_logger_redirect_test_/0-fun-67-'/0 (test/lager_test_backend.erl, line 1012)
**error:{assertEqual,[{module,lager_test_backend},
              {line,1012},
              {expression,"Level"},
              {expected,16},
              {value,8}]}
  output:<<"">>

  lager_test_backend: error_logger_redirect_test_ (warning reports are printed at the correct level)...*failed*
in function lager_test_backend:'-error_logger_redirect_test_/0-fun-68-'/2 (test/lager_test_backend.erl, line 1023)
in call from lager_test_backend:'-error_logger_redirect_test_/0-fun-71-'/0 (test/lager_test_backend.erl, line 1023)
**error:{assertEqual,[{module,lager_test_backend},
              {line,1023},
              {expression,"Level"},
              {expected,16},
              {value,8}]}
  output:<<"">>

  lager_test_backend: error_logger_redirect_test_ (single term warning reports are printed at the correct level)...*failed*
in function lager_test_backend:'-error_logger_redirect_test_/0-fun-72-'/2 (test/lager_test_backend.erl, line 1034)
in call from lager_test_backend:'-error_logger_redirect_test_/0-fun-75-'/0 (test/lager_test_backend.erl, line 1034)
**error:{assertEqual,[{module,lager_test_backend},
              {line,1034},
              {expression,"Level"},
              {expected,16},
              {value,8}]}
[...]

PR #328, opened 2016-03-11
Kill In The Event of Extreme Overload

blt:
This is a work-in-progress PR to integrate @jinnipark's work into mainline lager.  As of the issuing of this PR, the code does not fully work. In particular:

  * tests fail
  * restart of lager_event fails

This relates to Issue #324 (RIAK-1643) (RIAK-2068) (RIAK-2433).

### DIFF
--- a/src/lager_app.erl
+++ b/src/lager_app.erl
@@ -28,7 +28,8 @@
 -export([start/0,
          start/2,
          start_handler/3,
-         stop/1]).
+         stop/1,
+         boot/0]).
 
 -define(FILENAMES, '__lager_file_backend_filenames').
 -define(THROTTLE, lager_backend_throttle).
@@ -220,7 +221,10 @@ get_env_default({ok, Value}, _Default) ->
 
 start(_StartType, _StartArgs) ->
     {ok, Pid} = lager_sup:start_link(),
+    SavedHandlers = boot(),
+    {ok, Pid, SavedHandlers}.
 
+boot() ->
     %% Handle the default sink.
     determine_async_behavior(?DEFAULT_SINK,
                              application:get_env(lager, async_threshold),
@@ -246,8 +250,7 @@ start(_StartType, _StartArgs) ->
 
     clean_up_config_checks(),
 
-    {ok, Pid, SavedHandlers}.
-
+    SavedHandlers.
 
 stop(Handlers) ->
     lists:foreach(fun(Handler) ->
--- a/src/lager_app.erl
+++ b/src/lager_app.erl
@@ -146,24 +146,24 @@ start_error_logger_handler(_, HWM, {ok, WhiteList}) ->
                         throw({error, bad_config})
                 end,
 
-    KillerHWM = case application:get_env(lager, killer_hwm) of
-                    undefined -> -1;
-                    {ok, undefined} -> -1;
-                    {ok, KHWM} when is_integer(KHWM), KHWM >= 0 -> KHWM
-                end,
-    KillerReinstallAfter = case application:get_env(lager, killer_reinstall_after) of
-                               undefined -> 5000;
-                               {ok, undefined} -> 5000;
-                               {ok, V} when is_integer(V), V >= 0 -> V;
-                               {ok, BadKillerReinstallAfter} ->
-                                   error_logger:error_msg("Invalid value for 'cooldown': ~p~n",
-                                                          [BadKillerReinstallAfter]),
-                                   throw({error, bad_config})
-                           end,
-
-    _ = supervisor:start_child(lager_handler_watcher_sup,
-                               [?DEFAULT_SINK, lager_manager_killer,
-                                [KillerHWM, KillerReinstallAfter]]),
+    %% Conditionally start the high watermark killer.
+    case application:get_env(lager, killer_hwm) of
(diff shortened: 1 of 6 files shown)


jinnipark:
Thank you for this update.
I lost track of my PR for a while.
I'll close mine.

mrallen1:
This looks pretty good as a first pass. I am a little bit unsure how this will work for non-default sinks and it'd be nice to have a test case for that scenario. The timer:sleeps in the test code make me wary. I'd prefer to have something like

```erlang
wait_until(_Fun, 0) ->
    {error, too_many_retries};
wait_until(Fun, Retries) ->
     case Fun() of
         true -> ok;
         false -> timer:sleep(?SLEEP), wait_until(Fun, Retries-1)
     end.
```

blt:
Agreed, this is a pretty quick integration and does need to take into
account the other sinks.

Next step is getting the existing tests going, repairing the approach to be
inline with your previous and tackling additional sinks.

That seems straightforward. Present concern I have is that the sink doesn't
appear to come back online correctly. Did not have much investigation time.
Should be fun to figure out.

blt:
@jinnipark Thanks for doing the original work! :D

blt:
Hmm, on my local host two test suites--in addition to the tests that fail already on 3.1.0--still fail:

  * lager_manager_killer_test
  * lager_handler_watcher

Run in isolation, these suites run with no issue. I'll continue to look into the failure to try and pick apart what's going on, but if someone else had a notion to take a look I'd appreciate it. 

Question: is it expected that 8 tests in lager_test_backend will fail? I've confirmed failure on master branch from a clean build, on my system.

mrallen1:
If you're testing on erlang 18 then yes that's a known test fault we didn't fix because we're not regularly testing 18

There's a ticket open on this but I don't remember what it is.

blt:
One test failure remaining, tested on Erlang 17.

mrallen1:
For the record, #275 (RIAK-1886) is the known Erlang 18 test failures.

mrallen1:
Thanks for pulling this forward Brian. I'll take a look at it tonight or tomorrow.

blt:
Much appreciated! Going to get pulled off for a bit but I still have
ambitions to determine what's wrong with that final test.

mrallen1:
Your branch makes the reinstall fail because the handle_info which processes the `reinstall_handler` message was removed.  You can see on line 128 where a timer sends the message to the handler to reinstall a handler which failed to start initially.  This diff below makes all tests pass.

```diff
mallen-rMBP:lager mallen$ git diff
diff --git a/src/lager_handler_watcher.erl b/src/lager_handler_watcher.erl
index 336ee06..3d13cd3 100644
--- a/src/lager_handler_watcher.erl
+++ b/src/lager_handler_watcher.erl
@@ -80,6 +80,9 @@ handle_info({gen_event_EXIT, Module, Reason}, #state{module=Module,
         ok
     end,
     {noreply, State};
+handle_info(reinstall_handler, #state{module=Module, config=Config, sink=Sink} = State) ->
+    install_handler(Sink, Module, Config),
+    {noreply, State};
 handle_info(reboot, State) ->
     lager_app:boot(),
     {noreply, State};
```

blt:
Thanks @mrallen1!

mrallen1:
Cool. So the only remaining thing I'd like to see is a test with the killer that uses an alternative sink, then I think we'll be ready to merge this.

mrallen1:
btw - not necessarily nominating you to write that test @blt - just more of a TODO item for myself.

src/lager_app.erl
```diff
@@ -200,7 +221,10 @@ get_env_default({ok, Value}, _Default) ->
 
 start(_StartType, _StartArgs) ->
     {ok, Pid} = lager_sup:start_link(),
+    SavedHandlers = boot(),
+    {ok, Pid, SavedHandlers}.
 
+boot() ->
```
mrallen1: Let's make `boot/1` that takes a Sink argument. `boot/0` should call `boot/1` with the argument `?DEFAULT_SINK`

src/lager_manager_killer.erl
```diff
@@ -0,0 +1,51 @@
...
+
+-include("lager.hrl").
+
+-record(state, {
+          killer_hwm :: non_neg_integer(),
+          killer_reinstall_after :: non_neg_integer()
+         }).
+
+kill_me() ->
+    gen_event:call(lager_event, ?MODULE, kill_self).
```
mrallen1: This needs to be abstracted to use a generic Sink name.

mrallen1:
Just thought I'd update things. I spent most of yesterday and today working on this feature - generalizing it to multiple sinks, overhauling the start up code and so forth. I also spent some time with @macintux tracking down a race condition which revealed itself during testing. I'm currently seeing test failures in a single test and the failure is confounding me - I can't *quite* figure out why its failing, so I'm going to take a look at it tomorrow hopefully with fresh eyes and see if I can't figure it out.

When I get the last test functional I'm going to open a new PR which incorporates most of this PR and supersedes it.

blt:
Awesome, thank you for doing the work to push this over the line! This is
going to be really great.

mrallen1:
Closing in favor of #346

Issue #332, opened 2016-03-23
error_logger_lager_h:format_offender/1 looks for wrong key for offender name? [JIRA: RIAK-2471]


kennethlakin:
For ages, I've been getting Lager crash reports of the form "Supervisor $SUP had child undefined started with ...", regardless of whether or not the crashing child was a registered process. So, I did some digging.

At https://github.com/basho/lager/blob/master/src/error_logger_lager_h.erl#L298 , we see that the offender's reported Name is being sourced from the ``'name'`` key in the ``Off`` proplist. However, inspection of the contents of the proplist reveals that there is no ``'name'`` key, but there *is* an ``'id'`` key that contains what appears to the be child ID that was passed to the supervisor for the crashing process. When I make the obvious change the crash message changes to "Supervisor $SUP had child $CHILD_ID started with ...", which looks more reasonable to me.

Issue #14  looks vaguely related, though 6fdbe250718c510e53b5824a37dfeece330c94f0 reveals that the code in question *never* got the "name" from the offender's ``id`` key... Is this a long-standing typo and the intention was for format_offender to get the child's ID from ``id``, or is there something I'm missing here?

mrallen1:
Thanks for the report and thank you also for the code archeology. I can't speak to the history behind the changes, but I agree it would be nice to use the `id` if its available and has a value.  I think the way to do that would be to handle the case where name is `undefined` in a new function and return a different string more in line with your suggestion if `id` is populated.

kennethlakin:
Does PR #333  address this issue to your satisfaction? It seemed natural to me to leave the string as-is (because what it says makes a lot of sense, given the context), but just use ``id`` as the source of the child name if ``name`` is missing. If both are missing, we fall back to ``undefined``.

I ran the test suite before and after making the change. The same nine tests failed each time.

If you like, I can do more spelunking to determine where the ``name`` key gets set (if it *ever* does) to make sure that just using ``id`` (rather than making a more substantial change) is acceptable in this case.

macintux:
Digging through the history of the `supervisor` module it's clear that `name` was changed to `id` for the output of `extract_child/1` with this commit:

```
commit 1d4379e307cd7256e019e962b0ece6a1badcffb9
Author: Siri Hansen <siri@erlang.org>
Date:   Mon Jun 30 11:27:46 2014 +0200
```

Going to create a tweaked version of the PR that tries `id` first since that's the modern version of `name`. Thanks so much for investigating this and supplying a PR.

[...]

PR #333, opened 2016-03-23
Addresses issue #332.

kennethlakin:
This appears to be one way to address Issue #332 (RIAK-2471)

PR #334, opened 2016-03-25
Address #332: try both id and name

macintux:
The `stdlib` module `supervisor` reports the child name as `id` now vs `name` previously

### DIFF
--- a/src/error_logger_lager_h.erl
+++ b/src/error_logger_lager_h.erl
@@ -295,7 +295,10 @@ format_offender(Off) ->
         MFArgs ->
             %% regular supervisor
             MFA = format_mfa(MFArgs),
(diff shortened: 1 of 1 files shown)


macintux:
Supersedes #333. Thanks again @kennethlakin.

kennethlakin:
Formatting quibble: Lower down in the file, we that the second line of case assignments is indented by four spaces. This isn't how *my* editor is set up, and it's clearly how yours is either, but it *is* kinda nice.

mrallen1:
:+1: 0961013

mrallen1:
@borshop retry;

mrallen1:
:+1: 0961013

PR #315, opened 2016-01-20
Declare trace_files in configuration

nmichel:
Allow to declare trace_files directly in configuration. E.g. : 

```erlang
{lager, [
         {handlers, [
                     {lager_console_backend, info},
                     {lager_file_backend, [{file, "error.log"}, {level, error}]},
                     {lager_file_backend, [{file, "console.log"}, {level, info}]}
                    ]},

         {trace_files, [
                        {"log/events.log", [{module, mymodule}], info, [{formatter_config, [message, "\n"]}, {size, 1000}, {date, "$D0"}, {count, 5}]}
                       ]}
        ]},
```

PR #336, opened 2016-03-25
Ensure traces pointed to files work in config file

mrallen1:
Thanks to @nmichel - I wanted to handle traces defined in the configuration file consistently without a new keyword. So this is intended to supersede #315 but is *slightly* less functional because it doesn't allow the `Options` but on the other hand, it doesn't introduce a new configuration keyword either.

There is apparently a delay in flushing the log messages to files and so we have an evil timer:sleep/1 call in this test. I don't know of a good way to force the lager file backend to flush a file handle to disk on demand though. (Maybe it should be part of the API? Not sure right now.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -596,6 +596,9 @@ to `debug` as in the last file backend example.
 The `traces` keyword works on alternative sinks too but the same limitations
 and caveats noted above apply.
 
+**IMPORTANT**: You **must** define a severity level in all lager releases
+up to and including 3.1.0 or previous. The 2-tuple form wasn't added until
+3.2.0.
 
 Setting the truncation limit at compile-time
 --------------------------------------------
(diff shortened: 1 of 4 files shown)


mrallen1:
@borshop retry

mrallen1:
@macintux reminded me I need to add some updates to the README about this.

README.md
```diff
@@ -568,6 +568,35 @@ The former can be ameliorated by opening multiple traces; the latter
 can be fixed by rearchitecting lager's file backend, but this has not
 been tackled.
 
+### Traces from configuration
```
macintux: Since one of your examples (`event.log`) won't work with any existing release of lager, please add a note that the log level is required for the file backend in traces up to and including <fill-in-existing-2.x/3.x releases>.

mrallen1:
Good call. Added some language to that effect.

macintux:
:+1: 7aaa6ad

mrallen1:
@borshop merge

PR #337, opened 2016-03-29
Put the warning map value into process dictionary (Fix OTP 18 test failures)

mrallen1:
The sync_error_logger file looks up the message level with which to tag a message from a process dictionary entry named `warning_map'. When it is unset, it uses a default mapping which is fine for OTP 17 and before, but in OTP 18 and 19 the value changed from this default.

Now we put the current error logger warning mapping value into the process dictionary regardless of what OTP release tests are running under so that the messages do not cause failed tests if the default mismatches the OTP release default.

Addresses #275

### DIFF
--- a/test/lager_test_backend.erl
+++ b/test/lager_test_backend.erl
@@ -1128,8 +1128,9 @@ error_logger_redirect_test_() ->
             },
             {"warning messages with unicode characters in Args are printed",
                 fun(Sink) ->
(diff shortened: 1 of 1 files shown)


mrallen1:
@macintux I have added comments and updating the binding names like we discussed.

macintux:
:+1:

Issue #286, opened 2015-08-17
Dynamic sinks [JIRA: RIAK-2094]


walrusVision:
Currently all the extra sinks get configured on lager:start() and it pulls this information from the env via the config. I'd like to be able to configure sinks a bit more dynamically. E.g if someone starts an app and hasn't provided a config file then my app could configure the sink itself with some sensible defaults. Similar to how lager_app has a default configuration.

mrallen1:
Hi - thanks for the suggestion. It would be very difficult to dynamically start new sinks at run-time due to the way lager's parse transformation works.  The calls into lager's internal machinery for logging typically are done via this method at compile time.

It **is** possible to call the logger at run-time with an explicit sink of course using `lager:log/5` but it's two-times slower than the parse transform code. One way you could implement what you're asking for with respect to extra sinks with sensible defaults if none are given is to wrap the lager startup code with your own and inject the defaults into the application environment.

walrusVision:
Hello mrallen1. I have my apps parse transformed with sink names as they have the appropriate rebar settings. I am thinking about the case where the apps have been complied with parse transform, but a config defining the extra_sinks section hasn't been supplied so you get the {error,{sink_not_configured, Sink}} response when you try to log.

If I'm following this code correctly
https://github.com/basho/lager/blob/3.0.1/src/lager_app.erl#L214
https://github.com/basho/lager/blob/3.0.1/src/lager_app.erl#L167

Then at the moment lager just pulls the settings for the sinks from the environment config. If I have my parse_transform applied to the code already, would it then matter if the sink was configured on start, or at some latter point e.g. have my parse_transformed app call ```configure_sink\2``` it self with some settings?

walrusVision:
As a I test I exported ```lager_app:configure_sink\2``` so it was public. Then I tried these steps

 * Make a test lib which has a parse_transform in the rebar config for a extra sink called ```test_sink```, but it has no app config. It has one function which just does ```test_sink:info("Test message").```
 * Start lager
 * Start observer
 * Run my test function which calls the test sink and fails with sink_not_configured (which is expected)
 * Run configure_sink with just the settings I'd use in the app config normally
 * Watch the sink, links and handler appear in observer
 * The logging call in test function no longer fails and the message arrives at all the specified handlers correctly.

mrallen1:
OK. So the request is export `lager_app:configure_sink/2`?  That will meet your use case needs?

walrusVision:
[...]

PR #342, opened 2016-03-30
Export configure_sink/2 for runtime sink config

mrallen1:
Address #286 (RIAK-2094). This will enable applications to
dynamically configure log sinks, provided they
have already instrumented the code via parse
transformation to use those sinks.


macintux:
:+1: 1b582af

macintux:
@borshop merge

Issue #291, opened 2015-09-01
lager:log dialyzer error on level 'none' [JIRA: RIAK-2156]


walrusVision:
Dialyzer complains when you try to pass the lager:log functions a logging level of none.

e.g. 
```
ok = lager:log(
    Sink, none, self(),
    "Test none for ~p", [Sink]
  )
```

I assume that either `-type log_level()` should be changed to include it, or the spec on the log functions (log, dispatch_log) should be changed to `log_level() | none`

PR #343, opened 2016-03-30
Include 'none' in the log_level type

mrallen1:
Addresses #291 (RIAK-2156). External consumers who use 'none' as a level in their log messages will fail dialyzer.


macintux:
:+1: 5e6e396

mrallen1:
@borshop merge

PR #344, opened 2016-03-30
Flappy test clean up for TravisCI

mrallen1:
* filtermap - if we're going to have R15 as a Travis tested platform, this is required to ensure the implementation of the rotate sink feature does not break. (includes a dialyzer fixup)

* replace timer:sleep with wait_until

* increase timeout on gh280 test - 30 second for eunit, 10 seconds to receive the message


macintux:
:+1: 1453104

mrallen1:
@borshop merge

PR #355, opened 2016-05-10
Fix get_env bug

juise:
Fix for https://github.com/basho/lager/issues/354

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -40,9 +40,12 @@
 
 {eunit_opts, [verbose]}.
 {eunit_compile_opts, [
+    export_all,
+
     nowarn_untyped_record,
     nowarn_export_all
 ]}.
+
 {deps, [
     {goldrush, ".*", {git, "git://github.com/DeadZen/goldrush.git", {tag, "0.1.8"}}}
 ]}.
@@ -53,4 +56,3 @@
 {cover_enabled, true}.
 {edoc_opts, [{stylesheet_file, "./priv/edoc.css"}]}.
 
-{eunit_opts, [verbose]}.
--- a/src/lager_app.erl
+++ b/src/lager_app.erl
@@ -138,11 +138,10 @@ maybe_install_sink_killer(_Sink, HWM, ReinstallTimer) ->
     error_logger:error_msg("Invalid value for 'killer_hwm': ~p or 'killer_reinstall_after': ~p", [HWM, ReinstallTimer]),
     throw({error, bad_config}).
 
-start_error_logger_handler({ok, false}, _HWM, _Whitelist) ->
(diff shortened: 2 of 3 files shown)


mrallen1:
Yep, that's definitely a problem. Thanks for the patch. Looks like there's some kind of whitespace or indentation issue on line 240 and 242. Would you mind cleaning that up, please?

juise:
I'm do it intentionally, I think it's makes code more visually declarative. But if you think, that is really identation issue, i'll fix it!

And I wonder, why there is no test cases for this functionality?

mrallen1:
I would be grateful if you write some tests for this, yes.

juise:
Sure! I've write simple unit test for get_env/2,3 and get_env_default function. But the better way will be if you or someone who understand lager internals writes integration test (common test, not unit test) for lager application run, or for boot/0 function. Unfortunately I have no free time for it.

mrallen1:
Cool, thanks!

macintux:
Travis is showing more test failures than I'm seeing, but definitely there's some problem with the `lager_handler_watcher` tests.

```
module 'lager_handler_watcher'
  lager_handler_watcher: reinstall_on_initial_failure_test_...[6.079 s] ok
  lager_handler_watcher: reinstall_on_runtime_failure_test_...*failed*
in function lager_handler_watcher:'-reinstall_on_runtime_failure_test_/0-fun-4-'/0 (src/lager_handler_watcher.erl, line 183)
**error:{badmatch,undefined}
```

From running the test line by line it appears that the two "blind" pops are the problem: we're no longer getting an application startup or a gen_event crash message before the crash backend message.

mrallen1:
Yeah that was one of the failures that cropped up when we added the sink killer - I thought I had it fixed but obviously not.

mrallen1:
I just had an epiphany that maybe the reason it "works" on master is because this fix is the root cause of the bugs I encountered before.  I.e., they crash on master because this isn't fixed.

mrallen1:
Sorry I'm being imprecise. `lager_watcher_handler` crashes on master because this isn't fixed on master. Here it *doesn't* crash because this patch fixes the problem. So I am wondering if the code for the original test (like from 3.0.2 tag) works now.

macintux:
That’s quite possible. Will keep digging.

macintux:
Received independent confirmation of this; someone reported this problem with phoenix, ejabberd and lager. This patch fixed it.

I still haven't figured out what's happening to the lager start log message (wrt the failing test) but I'm confident this is the right fix. Will file another PR to address the weird whitespace issues and the failing test at a later point in time.

Since Bors won't merge this for us with a failing test, will do it manually.

:+1: 09aee4b

mrallen1:
Yeah I got an email too.

PR #359, opened 2016-05-19
add new log levels: all, trace

alexshavelev:


mrallen1:
Really strong :-1: on this without tests and a rationale. What problem are you trying to solve?

PR #360, opened 2016-05-19
Cleanup for test, whitespace

macintux:
* Undo unnecessary whitespace
* Make lager_handler_watcher test more resilient by scanning for
  messages of interest rather than assuming we know exactly what
  messages will arrive at the test backend
* `export_all` to allow R15 to run the `lager_app_tests` tests

### DIFF
--- a/src/lager_handler_watcher.erl
+++ b/src/lager_handler_watcher.erl
@@ -178,12 +178,10 @@ reinstall_on_runtime_failure_test_() ->
                     try
                         ?assert(lists:member(lager_crash_backend, gen_event:which_handlers(lager_event))),
                         timer:sleep(6000),
-                        _ = lager_test_backend:pop(), %% throw away application start up message
-                        _ = lager_test_backend:pop(), %% throw away gen_event crash message
-                        {_Severity, _Date, Msg, _Metadata} = lager_test_backend:pop(),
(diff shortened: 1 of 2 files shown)


javajolt:
Testing...

javajolt:
Tests pass on R16B02-basho10 and 18.3.3

src/lager_handler_watcher.erl
```diff
@@ -194,5 +192,17 @@ reinstall_on_runtime_failure_test_() ->
         ]
     }.
 
+pop_until(String, Fun) ->
```
javajolt: Out of curiosity, what other funs might you pass in here?

javajolt:
:+1: 124ba14

src/lager_handler_watcher.erl
macintux: Well, I pass `lists:flatten` for one invocation, but wrap it with further logic (`string:substr`) for another, or do I misunderstand your question?

That's all I expect for now, although the pattern seems generally useful.

javajolt: Yes, that was the question. `lists:flatten` makes a lot of sense.  The others might depend on use case. Thanks.

macintux:
@borshop merge

mrallen1:
Nice work 🍬

PR #365, opened 2016-06-28
Update TravisCI for OTP 18.3 and 19.0

mrallen1:
TravisCI now supports testing on 18.3 and 19.0. This PR updates the `.travis.yml` file to run tests against those OTP releases.

Switching to 18.3 instead of 18.2.1 makes sense as that's the release most will be targeting for 18.x


Vagabond:
sgtm.

PR #366, opened 2016-07-08
log the ranch error msg

getong:


mrallen1:
Thanks this is a good suggestion.

mrallen1:
We will be triaging PRs tomorrow, so I expect to merge soon. Thanks.

PR #367, opened 2016-07-14
Less lines on console

00dingens:
I output the lager logs on console and it is flooded with poor formatted outputs
Formatting with `~p` tries to break after 80 chars, so my output looked like this:

```
09:07:23.201 [error] Lager failed to install handler lager_journald_backend into lager_event, retrying later : {'EXIT',
                                                                                       {undef,
                                                                                        [{lager_journald_backend,
                                                                                          init,
                                                                                          [[{level,
                                                                                             debug}]],
                                                                                          []},
                                                                                         {gen_event,
                                                                                          server_add_handler,
                                                                                          4,
                                                                                          [{file,
                                                                                            "gen_event.erl"},
                                                                                           {line,
                                                                                            434}]},
                                                                                         {gen_event,
                                                                                          handle_msg,
                                                                                          5,
                                                                                          [{file,
                                                                                            "gen_event.erl"},
                                                                                           {line,
                                                                                            279}]},
                                                                                         {proc_lib,
                                                                                          init_p_do_apply,
                                                                                          3,
                                                                                          [{file,
                                                                                            "proc_lib.erl"},
                                                                                           {line,
                                                                                            240}]}]}}
```

This commit changes it to 

```
09:18:08.396 [error] Lager failed to install handler lager_journald_backend into lager_event, retrying later :
  {'EXIT',{undef,[{lager_journald_backend,init,[[{level,debug}]],[]},
                  {gen_event,server_add_handler,4,[{file,"gen_event.erl"},{line,434}]},
                  {gen_event,handle_msg,5,[{file,"gen_event.erl"},{line,279}]},
                  {proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,240}]}]}}
```

which does not flood the console so fast.

### DIFF
--- a/src/lager_handler_watcher.erl
+++ b/src/lager_handler_watcher.erl
@@ -128,7 +128,7 @@ install_handler2(Sink, Module, Config) ->
         Error ->
             %% try to reinstall it later
             ?INT_LOG(error, "Lager failed to install handler ~p into"
-               " ~p, retrying later : ~p", [Module, Sink, Error]),
+               " ~p, retrying later :~n  ~100p", [Module, Sink, Error]),
             erlang:send_after(5000, self(), reinstall_handler),
             ok
     end.


00dingens:
If the newline is not wanted then we need a higher number, like `~230p` .

mrallen1:
I'm sympathetic to this change, but I'm wary of using 100 columns. Can we make it "~n~p" and add a comment that we're using a newline so the stack trace is easier to read?

mrallen1:
@Vagabond pointed out that lager has a formatter for this kind of message already which is in `error_logger_lager_h:format_reason/1` so we should use that.

mrallen1:
I'm going to close this PR due to lack of activity.  Thanks for the patch.

PR #374, opened 2016-09-22
Restore backwards-compatible `lager_file_backend` configurations.

JeetKunDoug:
The strict configuration work in 31cf6c0 prevents some
backwards-compatible code in lager_file_backend from working as the
configuration is rejected before it gets to the backend. This fix adds
an additional guard on `check_handler_config/2` to support the `tuple`
configuration structure supported by `lager_file_backend`.


macintux:
+1 fce24bc

JeetKunDoug:
@borshop merge

PR #376, opened 2016-09-28
Fix dialyzer warnings

priestjim:
Add `| undefined` to records that get initialized with not all their set to the proper type (hence get the value undefined). These allow lager to exist as a dependency in a project and be filtered through that project's dialyzer run without warnings.

### DIFF
--- a/src/lager_crash_log.erl
+++ b/src/lager_crash_log.erl
@@ -46,8 +46,8 @@
 
 -record(state, {
         name :: string(),
-        fd :: pid(),
-        inode :: integer(),
+        fd :: pid() | undefined,
+        inode :: integer() | undefined,
         fmtmaxbytes :: integer(),
         size :: integer(),
         date :: undefined | string(),
@@ -119,7 +119,7 @@ schedule_rotation(Date) ->
 %% ===== Begin code lifted from riak_err =====
 
 -spec limited_fmt(string(), list(), integer()) -> iolist().
-%% @doc Format Fmt and Args similar to what io_lib:format/2 does but with 
+%% @doc Format Fmt and Args similar to what io_lib:format/2 does but with
 %%      limits on how large the formatted string may be.
 %%
 %% If the Args list's size is larger than TermMaxSize, then the
--- a/src/lager_file_backend.erl
+++ b/src/lager_file_backend.erl
@@ -57,8 +57,8 @@
 -record(state, {
         name :: string(),
         level :: {'mask', integer()},
-        fd :: file:io_device(),
-        inode :: integer(),
+        fd :: file:io_device() | undefined,
+        inode :: integer() | undefined,
         flap=false :: boolean(),
         size = 0 :: integer(),
         date :: undefined | string(),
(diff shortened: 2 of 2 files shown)


nickelization:
According to the Erlang docs, "for [record] fields without initial values, the singleton type 'undefined' is added to all declared types." http://erlang.org/doc/reference_manual/typespec.html#id79711

If I'm understanding this correctly, it seems like it should make no difference to these type specs whether we explicitly add 'undefined' to the specs or not.

On a fresh clone of lager, "make dialyzer" passes for me with no errors. Could you provide more info on what sort of errors you're seeing? I don't think I can figure out what kind of problems this PR would fix (other than the fix to the spec for `lager_file_backend:init`, which is clearly a good change).

priestjim:
I expected this behavior too but every time I include lager as a dependency and run dialyzer in my project I get warnings in the files I updated in this PR such as:

```erlang
_build/default/lib/lager/src/lager_crash_log.erl
  79: Record construction #state{name::atom() | binary() | [atom() | [any()] | char()],fd::'undefined',inode::'undefined',flap::'true'} violates the declared type of field fd::pid() and inode::integer()

_build/default/lib/lager/src/lager_file_backend.erl
  84: Invalid type specification for function lager_file_backend:init/1. The success typing is (maybe_improper_list() | {maybe_improper_list(),atom()} | {maybe_improper_list(),atom(),_,_,_}) -> {'error',{'fatal','bad_config'}}
 112: Record construction #state{fd::'undefined',inode::'undefined',flap::'false',date::'undefined' | [{'date',_} | {'day',integer()} | {'hour',_}],shaper::#lager_shaper{hwm::'undefined' | pos_integer(),mps::0,lasttime::{non_neg_integer(),non_neg_integer(),non_neg_integer()},dropped::0},last_check::{non_neg_integer(),non_neg_integer(),non_neg_integer()}} violates the declared type of field fd::pid() | {'file_descriptor',atom(),_} and inode::integer()
```

nickelization:
Huh, interesting. Thanks for the additional info @priestjim.

So @paegun did some additional testing on this and found that the dialyzer behavior seems to have changed somewhere between Erlang R16 and 19. Since the docs seem pretty explicit about how record type specs are supposed to work, I am suspicious that this might actually be a regression within dialyzer.

As soon as I get a chance, I'll post something to the erlang-users mailing list and see what people say. If this is the way dialyzer is supposed to work now, then I will happily merge your PR, but I just want to check and see what the story is behind this first :)

nickelization:
Okay, scratch that! @JeetKunDoug found a commit in OTP that explains this change in behavior: https://github.com/erlang/otp/commit/8ce35b287fb50a6845fccf6a13c672aae303dc91

So given the new behavior, these warnings actually make sense: those fields that dialyzer is complaining about really *should* have 'undefined' explicitly stated as part of the type spec, because sometimes we don't initialize them when creating a new state record.

I'm just going to run some smoke tests locally to double check that we haven't missed anything, but assuming that all looks good I will approve and merge this PR.

nickelization:
:+1: 7598d8c

nickelization:
@priestjim Thanks for the contribution!

PR #377, opened 2016-09-29
Support gen_statem in error logger

priestjim:
Add a simple case to support gen_statem crashes, since they generate the
same message in error_logger but not the same returned state.

### DIFF
--- a/src/error_logger_lager_h.erl
+++ b/src/error_logger_lager_h.erl
@@ -33,7 +33,7 @@
 
 -export([format_reason/1, format_mfa/1, format_args/3]).
 
--record(state, { 
+-record(state, {
         sink :: atom(),
         shaper :: lager_shaper(),
         %% group leader strategy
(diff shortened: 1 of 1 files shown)


pj-splunk:
@nickelization care to take a look on this one as well?

mrallen1:
This is really nice. Thank you!

Would you mind adding a test [here](https://github.com/erlang-lager/lager/blob/master/test/lager_test_backend.erl#L902) so I can get this merged?

priestjim:
@mrallen1 how should the test be implemented? It feels like it's just a case of gen_fsm messages vs gen_statem messages and gen_fsm crash capturing is not really tested in lager

mrallen1:
You know, you're right! I had thought we covered it, but we haven't!

mrallen1:
I pushed a new branch `fsm_test` which has a gen_fsm and gen_statem that crash and conditionally compile statem on OTP 19+ only. So my plan is to merge your PR into that branch and then put at least one simple test in.

mrallen1:
I'm going to close this now as I have green tests on #391. Thanks for your patch!

PR #380, opened 2016-11-28
Allow Metadata-only logging

varnerac:
This allows users to log metadata (tuple list) without a format string in lager. For
example `lager:info([{foo, bar}])` instead of `lager:info([{foo,bar}], "")`.


mrallen1:
Thanks.

PR #385, opened 2017-01-20
Fix async_threshold_test_ and trace_files_test_ flappiness.

tburghart:
This PR extends the changes in https://github.com/erlang-lager/lager/pull/384 - it ***may*** be mergeable separately, but has not been tested independently.

Initially, problems with `async_threshold_test_` were thought to be related to newer versions of OTP _and/or_ newer multicore CPUs handling messages more efficiently, thereby preventing automatic transition to synchronous `lager` message handling.

After many experimental changes, it appears that the root problem was pollution of the test environment by previous external tests that don't clean up after themselves properly, so this test now rigorously scrubs its environment before running.  The changes already in place are deemed worthwhile and have been kept.

Timeout issues with `trace_files_test_`, where we don't know how long it will take for updates to be flushed, are also addressed.

### DIFF
--- a/src/lager_crash_log.erl
+++ b/src/lager_crash_log.erl
@@ -1,4 +1,6 @@
-%% Copyright (c) 2011-2012 Basho Technologies, Inc.  All Rights Reserved.
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2011-2017 Basho Technologies, Inc.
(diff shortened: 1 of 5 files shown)


mrallen1:
Thanks for the contribution. This race condition is one that we're wrestled with for a while.

PR #386, opened 2017-01-30
Fix path expansion for traces when log_root is set

hcs42:
Previously the following configuration didn't work properly:

    {log_root, "logs"},
    {handlers, [{lager_file_backend, [{file, "access.log"}, {level, none}, ...]}]},
    {traces, [{{lager_file_backend, "access.log"}, [{tag, access}], info}]}

When lager:trace_file was trying to find a file among the existing handlers, it was searching for the expanded file name ("logs/access.log"), but the handlers store the file names without expansion ("access.log"), so it didn't find it.

The result was that lager:trace_file started another "access.log" backend, which of course didn't use e.g. the log rotation settings of the first handler. When the log_root was relative, then each log message appeared twice in the log file (both backends thought that they should log it, because when calculating the Destination, "logs/access.log" was used in both cases).

Now the code performs path expansion on the file name inside the handler too. Also, now it calculates the absolute paths of both files, so that if one file is specified with an absolute path and the other with a relative path, lager will still know that they are the same.

### DIFF
--- a/src/lager.erl
+++ b/src/lager.erl
@@ -214,7 +214,7 @@ trace_file(File, Filter, Level, Options) ->
         {Sink, {ok, Trace}} ->
             Handlers = lager_config:global_get(handlers, []),
             %% check if this file backend is already installed
-            Res = case lists:keyfind({lager_file_backend, FileName}, 1, Handlers) of
+            Res = case lager_util:find_file(FileName, Handlers) of
                       false ->
                           %% install the handler
                           LogFileConfig =
--- a/src/lager_util.erl
+++ b/src/lager_util.erl
@@ -28,7 +28,7 @@
     open_logfile/2, ensure_logfile/4, rotate_logfile/2, format_time/0, format_time/1,
     localtime_ms/0, localtime_ms/1, maybe_utc/1, parse_rotation_date_spec/1,
     calculate_next_rotation/1, validate_trace/1, check_traces/4, is_loggable/3,
-    trace_filter/1, trace_filter/2, expand_path/1, check_hwm/1,
+    trace_filter/1, trace_filter/2, expand_path/1, find_file/2, check_hwm/1,
     make_internal_sink_name/1, otp_version/0
 ]).
 
@@ -513,6 +513,26 @@ expand_path(RelPath) ->
         undefined -> % No log_root given, keep relative path
             RelPath
     end.
+
+%% Find a file among the already installed handlers.
+%%
+%% The file is already expanded (i.e. lager_util:expand_path already added the
+%% "log_root"), but the file paths inside Handlers are not.
+find_file(_File1, _Handlers = []) ->
+    false;
+find_file(File1, [{{lager_file_backend, File2}, _Handler, _Sink} = HandlerInfo | Handlers]) ->
+    File1Abs = filename:absname(File1),
+    File2Abs = filename:absname(lager_util:expand_path(File2)),
+    case File1Abs =:= File2Abs of
+        true ->
+            % The file inside HandlerInfo is the same as the file we are looking
+            % for, so we are done.
+            HandlerInfo;
+        false ->
+            find_file(File1, Handlers)
+    end;
+find_file(File1, [_HandlerInfo | Handlers]) ->
+    find_file(File1, Handlers).
 
 %% Log rate limit, i.e. high water mark for incoming messages
 
--- a/test/lager_trace_test.erl
+++ b/test/lager_trace_test.erl
@@ -1,0 +1,103 @@
+-module(lager_trace_test).
+
+-compile([{parse_transform, lager_transform}]).
+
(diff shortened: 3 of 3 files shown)


mrallen1:
Thanks, this is super dope. I have one small request, though. Would you please add a few test cases for your code so we can be sure we don't regress on this in the future?

hcs42:
Sure!

hcs42:
Update: witing the automatic test is in progress (it's a bit more tricky than I expected :) )

mrallen1:
It can be. If you're stuck, I'd be happy to help point you in the right direction. Thanks so much for your contribution.

hcs42:
Sorry, I've been busy with other things... I will look at this on the weekend and either finish the test (and rebase it :) ) or ask for some pointers.

mrallen1:
Thanks!

hcs42:
I have added the test and force pushed the commit.

The test contains a timer:sleep(1000) call, because the error with the old code is that *eventually* the log file will contain the log messag twice. My idea for testing this *eventually* is to sleep one second and check if duplication has happened. If there is a better way, I'm open to suggestions.

mrallen1:
This is great. Curious why you didn't enclose the timer in the wait_for function you quoted - that seems like a better way to handle that "wait for" condition than a blanket sleep. If you want to adjust the time between the first call and the second, just make the sleep time a parameter you pass into it.

hcs42:
Thanks for the tips!

> Curious why you didn't enclose the timer in the wait_for function you quoted - that seems like a better way to handle that "wait for" condition than a blanket sleep.

Simply enclosing the timer in wait_for (or wait_until, as it's more often seems to be called in the codebase) would make it easy for the test to pass even when it is buggy. After the first line is written into the file, wait_for could quickly conclude that this is fine, test passed. But it may happen that the second line is written into the file a little later. (This is what I usually saw in the production log files.) My solution was make sure that we give time to the system for the second line to arrive, hence timer:sleep(1000). I could add a wait_for call after the timer:sleep(1000), so that if the first line arrives later than a second, we still pass.

> If you want to adjust the time between the first call and the second, just make the sleep time a parameter you pass into it.

If we want to replace timer:sleep with wait_for, we need to make sure that the sleep happens first (currently wait_until is happy to continue quickly if the condition holds). So we would need to move the timer:sleep inside wait_for to the beginning.

What do you think, does it make sense to add a wait_for after the timer:sleep(1000), or to create a new wait_for with the sleep parameter and moving the timer:sleep to the beginning?

mrallen1:
I would wrap the rest of the test (after waiting for the file to exist) in another wait_for function. I would simplify the output of lager's messages using a custom formatting function. Something like

```erlang
format(Msg, _Config) -> lager_msg:message(Msg)
```

should get you the "bare" message contents without timestamps or severity, etc. You can use this format/2 function as a "custom formatter" when you set up your test file in the test setup.  (See [Custom Formatting](https://github.com/erlang-lager/lager#custom-formatting) in the README.)

Then I would repeatedly read the log file contents, split it into lines with something like

```erlang
binary:split(FileContents, <<"\n">>, [global, trim_all])
```

and inspect the list that results to see if your log message has been duplicated. 

If not, then wait some more. If the log message does not show up as duplicated after the wait_for times out, then pass the test, otherwise, fail the test if a duplicate message shows up.

Since this test might run for 2 or 3 seconds, it would be a good idea to wrap the whole thing in an extended timeout tuple for EUnit.

```
{timeout, 10, fun do_test/0}
```

or whatever works best.  Running unit tests that go quick on your own workstation take 2 or 3 times longer on TravisCI and can lead to spurious test failures.

mrallen1:
During our lager PR triage today, @jlouis suggested a function like the one in this gist might be a good way to build a test that waits for a condition with a sleep that backs off exponentially

https://gist.github.com/anonymous/a44fefd9dbf6d610b2be9aa7ec7079d0

Would you have time to incorporate something like that in your test?

hcs42:
Thank you. I will do that on the weekend.

hcs42:
Thank you for all your suggestions (and patience...)

I incorporated all of them. Feel free to modify it before merging, especially the timeout values.

* I use another `wait_until` function to check for the duplicate log entry.
* I use lager's formatter to make the message minimal.
* I use binary:split on the file contents. I actually only examine the number of lines when everything looks ok.
* I wrapped the unit test into a timeout tuple.
* I used jlouis's gist to back off according to the Fibonacci sequence.

hcs42:
Note: I used `trim` instead of `trim_all` in `binary:split`. The latter works only from Erlang 18, which made travis unhappy.

mrallen1:
Looks really good. I just got back from traveling, so I'm going to wait until tomorrow to merge but, this is fantastic. 

Thank you!

mrallen1:
Just ran everything locally to confirm. This looks good. Thanks!!

PR #387, opened 2017-01-31
Make "it" unambiguous

hcs42:
It wasn't clear whether "it" referred to async_threshold or async_threshold_window.


mrallen1:
Thanks!

PR #391, opened 2017-02-24
Format gen_fsm/gen_statem crashes correctly

mrallen1:
Based on and supersedes #377.

Includes unit tests for the crash messages and formatting for both gen_fsm and gen_statem.


Vagabond:
+1

PR #392, opened 2017-02-27
Support for dynamic metadata

walrusVision:
This is a patch to support metadata key/value pair which returns a dynamic value. This is to provide functionality akin to Python's [LoggerAdpator](https://docs.python.org/2/howto/logging-cookbook.html#context-info) 

* Metadata can either be provided at compile time via the parse_transform or per-process when added using `lager:md/1`.
* Metadata functions are only evaluated on message emit
* By default there is no change to existing behavior, except in the case of somebody wanting to log a function reference without converting it first, e.g if you wanted to see `#Fun<module.function.arity>` in the logs, however all tests currently pass
* Have written some tests to ensure the correct behavior of this new behavior


README.md
```diff
@@ -636,6 +636,29 @@ and caveats noted above apply.
...
+              [{metadata_name, {module_name, function_name}]}.
+```
+
+The MF called should take no arguments and should return a value that can be be
+ formatted into a log message. Any function which is not defined, loaded, or
+ throws an exception will return `undefined` or the provided default value.
+
+This metadata is also persistent across processes.
+ 
+**IMPORTANT**: Since you can calling functions, be aware that using these
```
mrallen1: Suggest an edit like

Since this feature relies on function calls injected at the point where a log may or may not be generated, your logging performance (ops/sec) will be impacted by what the functions you call do and how much latency they may introduce.

test/lager_test_function_transform.erl
```diff
@@ -0,0 +1,152 @@
...
+]).
+-endif.
+
+-ifdef(TEST).
+
+transform_static() ->
+  static_result.
+
+transform_dynamic() ->
+  erlang:monotonic_time().
```
mrallen1: This is going to make all tests that aren't running OTP 19 fail as Travis shows.  One solution would be to use `lager_util:otp_version/0` in your tests to make the appropriate call to the function that is correct depending on the OTP version.

walrusVision:
If you don't use this functionality there should be no noticeable effect on performance.

The only places where it has an impact will be in the formatter when trying to resolve an atom in the configuration which has to make a call to `get_metadata`. The function now has to evaluate a guard before falling through to the next case. 

Doing some basic tests of performance, doing lookup on the old and new `get_metadata` 10 million times takes about 3.9 seconds for each of them, with the new function being about 0.02 seconds slower on average, with 0.08 seconds being the worst and -0.02 being the best I saw.

mrallen1:
Thanks. We will discuss this change during our monthly triage session - scheduled for March 16, 2200 UTC (1400 US/Pacific)

walrusVision:
Hi @mrallen1, I've had to push another change, as I've just discovered a race condition. 

* Logger Process logs a message
* Lager Process transmits message through to the backend async and returns
* Function Process dies
* Lager Backend Process starts formatting message, can't resolve the function as the function has now died.
 
The change I've pushed is to allow the function to be resolve at log or on emit (and you can specify per function). While resolving on log won't completely eliminate this race condition, if the Logger Process and the Function Process are the same process/app then it should reduce it. I've updated the README with the consequences, advantages/disadvantages of either.

mrallen1:
Hey - thanks for the PR. We discussed it today and have a few questions:

* What's the use case for this functionality? Can you give a scenario where its useful for you?
* Why does the function not accept the metadata lager already injects (like pid, line number, etc)

Would it be possible to use the `lager:md/1` functions to achieve similar ends?

walrusVision:
Hi @mrallen1 

Thank you for taking the time to look at the pull request.

Our particular use case for this functionality is tracing calls across systems. For the Erlang parts we are attempting to use [sequential tracing](http://erlang.org/doc/man/seq_trace.html) to trace calls through various Erlang apps, so that when we log messages information relating to the call being traced is included. 

E.g. In our output log we can have this sort of format
2017-03-17 10:00:00.000 app1 <trace id 1> INFO <message>
2017-03-17 10:00:00.001 app1 <trace id 2> INFO <message>
2017-03-17 10:00:00.002 app2 <trace id 2> INFO <message>
2017-03-17 10:00:00.003 app2 <trace id 1> INFO <message>

If we had unlimited time we would integrate this with something like [Zipkin](http://zipkin.io), however in the short term it is unlikely.

Initially we looked at using lager:md to store the information, however the nature of sequential tracing would require us to call lager:md in every message handler that might directly or indirectly log a message, and also after we receive any reply message (e.g. by directly or indirectly invoking gen_server:call). This proved to be somewhat onerous and error prone, and having lager automatically call a function as part of logging seemed like a cleaner solution. The particular function we are using queries the sequential trace label for the calling process, and uses this to look up trace information to be added to the log message. This is why we had to define the on_log | on_emit syntax for the transform since the trace information could be gone by the time the message was emitted by the backend, and so needs to be queried at the point of logging. 

As for why the function doesn't accept the other populated values - we hadn't considered it as it wasn't required for our use case, however we can't think of any reason why it shouldn't.

walrusVision:
Just bumping this as reminder to discuss during the next monthly triage session

mrallen1:
Thanks for the bump. 

We did discuss it today and I did merge it onto a local branch to play around with it. 

Would you please extend your write up in the README to include some examples of the kind of output the `on_emit` and `on_log` functions should generate?  

I see in the test you're using a literal atom as an example of static data and in the other test function you're using `monotonic_time` - I think the feature write up in the README would be a lot clearer if you would put in some good examples and illustrate why someone might want to use this feature.

mrallen1:
Bump on this? Are you still interested in merging this?  Thanks.

walrusVision:
Hi @mrallen1, yes I'm still interested. I'll have some time soon to work on updating the README.

mrallen1:
Great, thank you!

mrallen1:
Thanks for the update!

PR #389, opened 2017-02-17
Add several new trace filters

mrallen1:
* Expose new trace filter functionality enabled by new versions of goldrush
* Enable filter expression composition using `any` or `all`
* Enable null filter expressions
* Update the README with these new expressions

Updates and supercedes #160, addresses #159

PR #396, opened 2017-03-09
This adds the ability to EXCLUDE from a trace

DeadZen:
Goldrush was updated in 3.2.3 to 0.19, This extends the feature added by it for use in lager.


DeadZen:
This is redundant with #389

Issue #382, opened 2017-01-14
Capture process lager_metadata in error_logger handler and pass it on as metadata to lager itself


Vagabond:
When a process that has set lager metadata crashes, we should, in the error_logger handler, capture that dictionary and pass it to lager again.

mrallen1:
See https://github.com/erlang-lager/lager/blob/master/src/error_logger_lager_h.erl#L55
we'd add metadata there beyond just the pid

Vagabond:
https://github.com/erlang-lager/lager/blob/master/src/error_logger_lager_h.erl#L307 actually implements this.

PR #397, opened 2017-03-13
Capture error-site metadata from error_logger messages & feed them to lager

Vagabond:
Previously, error_logger messages only had the pid of the crashed
process, not any of the traditional lager metadata like
module, function, line, etc.

It turns out, however, that we already have most if not all of that
information as metadata already in the stack trace. Thus, we can capture
it and re-inject the error logger log message as a lager message with
metadata.

Additionally, when a process crashes, we get the process dictionary
passed to us, which we can scan for the '_lager_metadata' entry, and
attach the dead process' metadata to the log message as well.

### DIFF
--- a/src/error_logger_lager_h.erl
+++ b/src/error_logger_lager_h.erl
@@ -148,20 +148,23 @@ log_event(Event, #state{sink=Sink} = State) ->
                     %% gen_server terminate
                     [Name, _Msg, _State, Reason] = Args,
                     ?CRASH_LOG(Event),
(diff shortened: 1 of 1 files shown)


mrallen1:
This looks great. Addresses #382 - looks like the one test failure is due to that old race condition with flushing startup messages and not due to this code change. One of these days I really need to kill that race condition properly.

👍

Vagabond:
If you are happy with this, please feel free to merge.

Issue #348, opened 2016-04-09
Add docs about using lager in elixir [JIRA: RIAK-2495]


mrallen1:
We should include in the readme a section about using lager in the context of elixir apps or logging abstraction libs like hut

zhaohansprt:
then  how it goes?

zhaohansprt:
i  found  there is  the way to  call  lager in elixir but seemed not like  in erlang
:lager.log(:error, self(), "message ~p",[:erlang.node()])

mrallen1:
Hi - I don't think anyone who works on lager day-to-day has the necessary experience with elixir to write a documentation section like this. But we would welcome someone who does to help with that section of the docs.  

That's why I tagged this issue as "help wanted"

holsee:
I'll give this a shot.  A few things worth figuring out first...

I would say as an elixir dev there would be a few options that come to mind when including lager in a project.

I would be asking myself:

- "Is there a wrapper that would allow for compile time macro benefits for log level exclusion?" 
- "is my logging going to become a risk / be impacted through using the elixir Logger abstraction with lager backend?"

Questions like this would be great to have an answer for.

So the first thing to think about would be to figure out if I wished to use lager as a [Logger backend](https://hexdocs.pm/logger/Logger.html#module-backends), or directly. If I opt for directly, I would then decide if I wanted to use a wrapper weighing up the usual pros and cons of doing such a thing.

#### Option 1: The Elixir Logging abstraction (with a lager backend)
see: https://hexdocs.pm/logger/Logger.html#module-backends

The elixir Logger is pretty nice, and allows for "backends" to be plugged in, whilst providing a nice idiomatic logging api. Other benefits of this approach come from the elixir macro system which enables caller introspection and the logging calls to be stripped out during compile time for the logging levels that the application doesn't wish to include for a given build type.

From what I remember backends are super simple to plug in, back pressure is handled quite nicely and the adapter is supervised in a standard OTP fashion.  

I would for sure be opting for this approach myself if I was, say, writing a library for redistribution in elixir and on the same note many elixir libs that I may be pulling in would also use the elixir `Logger`, so this scenario is certainly going to be common and should probably be mentioned.

Now wether or not lager contributors wish to provide a reference implementation is open to discussion.

#### Option 2: Using lager directly

[...]

PR #398, opened 2017-03-17
Added elixir section to README

holsee:
Documentation for Elixir usage of Lager. #348 

I aimed to provide the useful context that would allow elixir developers to make a choice between directly usage or the more idiomatic Lager as a Logger Backend approach. 

I felt it would be overkill to add elixir example for every Erlang snippet in the README.


mrallen1:
This is lovely, thank you. I'd like to add a quick section about [hut](https://github.com/tolbrino/hut) too but I can do that independently.

PR #404, opened 2017-04-13
Document placeholders introduced by error logger integration

lucafavatella:


lucafavatella:
I doubt I managed to break with this change the following test as reported by CI:
```
  lager_test_backend: async_threshold_test_ (async threshold works)...*failed*

in function gen:do_call/4 (gen.erl, line 219)

in call from gen_event:rpc/2 (gen_event.erl, line 189)

in call from lager_test_backend:'-async_threshold_test_/0-fun-10-'/0 (test/lager_test_backend.erl, line 1718)

**exit:killed
```

mrallen1:
Yes that test very occasionally has a spurious failure due to a race condition we can't eliminate 100%. I'm re-running the test suite.

mrallen1:
Thanks for the doc update.

Issue #168, opened 2013-08-24
Add an option for lager_console_backend for output to STDERR


joedevivo:
While this is being done, it would be a great opportunity to migrate lager_console_backend to the new style config.

jlouis:
Some investigation:

The console backend logs via `io:put_chars(user, ...)` (see `handle_event/2` in `lager_console_backend`). You can send data to stderr by using `standard_error` in place of `user`. But more work is needed in order to make this a configurable option, notably coming up with the necessary changes to the console backend configuration line. Perhaps you would need to simply accept:

    {lager_console_backend, info, stderr}

in lieu of

    {lager_console_backend, info}

mrallen1:
As #409 is merged and 3.5.0 released, closing this issue. Thanks!

PR #409, opened 2017-05-19
Enable sending console output to standard error

mrallen1:
Also rationalized the configuration options for the console backend and adds deprecation warnings for the older configuration styles.

TODO:
- [x] Tests

@Vagabond what would be a good way to test these changes?  We have a little pretend group leader shim in the unit tests. Can we use the same technique for `standard_error`

Addresses #168

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,29 @@ Features
 * Optional load shedding by setting a high water mark to kill (and reinstall)
   a sink after a configurable cool down timer
 
+Contributing
+------------
+We welcome contributions from the community. We are always excited to get ideas
+for improving lager.
+
+If you are looking for an idea to help out, please take a look at our open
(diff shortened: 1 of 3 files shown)


mrallen1:
OK, got some validation unit tests in there. When you get a chance, please give a look @Vagabond. Thanks!

pselden:
Can this be configured per log-level? Meaning, can I send info to stdout and error to stderr?

Vagabond:
I *think* you could do this by installing the 2 instances of the console handler, one logging to stderr with loglevel of '<error' and one logging to stderr with '>error'.

Actually, it doesn't look like we support more than one instance of the console_backend right now, we'd have to extend it to export `config_to_id/1` so they could have different names in the gen_event.

PR #410, opened 2017-05-27
Fixes for OTP 20

Vagabond:
* Handle new gen_server crash messages that contain client pid and stacktrace
* Update platform define for OTP 20
* Handle arity change in gen_fsm handle_msg


mrallen1:
Thanks! Will try to take a look tonight!

mrallen1:
Tests are green! Thanks!

PR #411, opened 2017-05-28
Make the high water mark shaper better

Vagabond:
* If the shaper is in overload and the final message comes in, but no
  further messages arrive for some time, until another message came in,
  the drop count would not be printed. Now we set a timer to ensure it
  prints the drop count after the current second expires.
* Allow the shaper to take a filter function that allows events that
  would not normally be printed anyway to not be counted against the
  HWM. This means that if you're suppressing supervisor startup messages
  you won't see drop events counted for messages you'd never see
  printed.

### DIFF
--- a/src/error_logger_lager_h.erl
+++ b/src/error_logger_lager_h.erl
@@ -88,29 +88,27 @@ shaper_fun() ->
         {false, false} ->
             fun(_) -> false end;
         {true, true} ->
-            fun({info_report, _GL, {_Pid, std_info, D}}) when is_list(D) ->
-                    lists:member({exited, stopped}, D);
-               ({info_report, _GL, {_P, progress, D}}) ->
-                    (lists:keymember(application, 1, D) andalso lists:keymember(started_at, 1, D)) orelse
-                    (lists:keymember(started, 1, D) andalso lists:keymember(supervisor, 1, D));
-               (_) ->
-                    false
-            end;
+            fun suppress_supervisor_start_and_application_start/1;
         {false, true} ->
-            fun({info_report, _GL, {_Pid, std_info, D}}) when is_list(D) ->
-                    lists:member({exited, stopped}, D);
-               ({info_report, _GL, {_P, progress, D}}) ->
-                    lists:keymember(application, 1, D) andalso lists:keymember(started_at, 1, D);
-               (_) ->
-                    false
-            end;
+            fun suppress_application_start/1;
         {true, false} ->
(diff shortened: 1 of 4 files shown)


src/error_logger_lager_h.erl
```diff
@@ -83,8 +83,37 @@ handle_call({set_high_water, N}, #state{shaper=Shaper} = State) ->
 handle_call(_Request, State) ->
     {ok, unknown_call, State}.
 
+shaper_fun() ->
+    case {lager_app:get_env(lager, suppress_supervisor_start_stop, false), lager_app:get_env(lager, suppress_application_start_stop, false)} of
+        {false, false} ->
+            fun(_) -> false end;
+        {true, true} ->
+            fun({info_report, _GL, {_Pid, std_info, D}}) when is_list(D) ->
+                    lists:member({exited, stopped}, D);
```
mrallen1: I would really like to see this anonymous fun in its own named function - I think it will make the code clearer and any stacktraces more obvious.

src/error_logger_lager_h.erl
```diff
@@ -83,8 +83,37 @@ handle_call({set_high_water, N}, #state{shaper=Shaper} = State) ->
...
+               ({info_report, _GL, {_P, progress, D}}) ->
+                    (lists:keymember(application, 1, D) andalso lists:keymember(started_at, 1, D)) orelse
+                    (lists:keymember(started, 1, D) andalso lists:keymember(supervisor, 1, D));
+               (_) ->
+                    false
+            end;
+        {false, true} ->
+            fun({info_report, _GL, {_Pid, std_info, D}}) when is_list(D) ->
```
mrallen1: Same comment as above. Would you please make this its own named function?

src/error_logger_lager_h.erl
```diff
@@ -83,8 +83,37 @@ handle_call({set_high_water, N}, #state{shaper=Shaper} = State) ->
...
+                    lists:member({exited, stopped}, D);
+               ({info_report, _GL, {_P, progress, D}}) ->
+                    lists:keymember(application, 1, D) andalso lists:keymember(started_at, 1, D);
+               (_) ->
+                    false
+            end;
+        {true, false} ->
+            fun({info_report, _GL, {_P, progress, D}}) ->
```
mrallen1: Same here.

Vagabond:
Reworked as requested.

mrallen1:
👍 Thanks! I think this is great!

PR #412, opened 2017-05-29
Fix the bug with the wrong format.

lintingbin2009:
This is a little problem, it will be better to fix it.


mrallen1:
Good catch. Thanks!

PR #418, opened 2017-09-08
A little emphasis on application variable crash_log

suexcxine:
Emphasize application variable, since I missed it. 

A little emphasis may help.


mrallen1:
Thanks for the patch!

PR #419, opened 2017-09-17
support logging with macros per level

tsloughter:
With new macros being supported in OTP it makes sense to provide a macro interface to lager as well.

There are still some forms of these functions I need to support, like `?info_unsafe`? I stopped here to check that this was an acceptable change upstream and if so that the api was what you'd want.

### DIFF
--- a/include/lager.hrl
+++ b/include/lager.hrl
@@ -30,11 +30,11 @@
                            {line, ?LINE} | Extras]).
 
 -define(lager_log(Severity, Format, Args, Safety),
-        ?lager_log(?DEFAULT_SINK, Severity, ?METADATA([]), Format, Args,
-             ?DEFAULT_TRUNCATION, Safety)).
+        ?lager_log(?DEFAULT_SINK, Severity, ?METADATA(lager:md()), Format, Args,
+                   ?DEFAULT_TRUNCATION, Safety)).
 -define(lager_log(Severity, Metadata, Format, Args, Safety),
-        ?lager_log(?DEFAULT_SINK, Severity, ?METADATA(Metadata), Format, Args,
-             ?DEFAULT_TRUNCATION, Safety)).
+        ?lager_log(?DEFAULT_SINK, Severity, ?METADATA(Metadata++lager:md()), Format, Args,
+                   ?DEFAULT_TRUNCATION, Safety)).
 
 -define(lager_log(Sink, Severity, Metadata, Format, Args, Size, Safety),
         lager:dispatch_log(Sink, Severity, Metadata, Format, Args, Size, Safety)).
--- a/include/lager.hrl
+++ b/include/lager.hrl
@@ -20,6 +20,48 @@
 -define(DEFAULT_SINK, lager_event).
 -define(ERROR_LOGGER_SINK, error_logger_lager_event).
 
+-define(METADATA(Extras), [{severity, info},
+                           {pid, self()},
+                           {node, node()},
+                           {module, ?MODULE},
+                           {function, ?FUNCTION_NAME},
+                           {function_arity, ?FUNCTION_ARITY},
+                           {file, ?FILE},
+                           {line, ?LINE} | Extras]).
+
+-define(log(Level, Format, Args, Safety),
+        lager:dispatch_log(?DEFAULT_SINK, Level, ?METADATA([]), Format, Args,
+                           ?DEFAULT_TRUNCATION, Safety)).
+-define(log(Level, Metadata, Format, Args, Safety),
+        lager:dispatch_log(?DEFAULT_SINK, Level, ?METADATA(Metadata), Format, Args,
+                           ?DEFAULT_TRUNCATION, Safety)).
+
+-define(debug(Format, Args), ?log(debug, Format, Args, safe)).
+-define(debug(Metadata, Format, Args), ?log(debug, Metadata, Format, Args, safe)).
+
+-define(info(Format, Args), ?log(info, Format, Args, safe)).
+-define(info(Metadata, Format, Args), ?log(info, Metadata, Format, Args, safe)).
+
+-define(notice(Format, Args), ?log(notice, Format, Args, safe)).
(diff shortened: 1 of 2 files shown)


include/lager.hrl
```diff
@@ -20,6 +20,48 @@
...
+                           {pid, self()},
+                           {node, node()},
+                           {module, ?MODULE},
+                           {function, ?FUNCTION_NAME},
+                           {function_arity, ?FUNCTION_ARITY},
+                           {file, ?FILE},
+                           {line, ?LINE} | Extras]).
+
+-define(log(Level, Format, Args, Safety),
+        lager:dispatch_log(?DEFAULT_SINK, Level, ?METADATA([]), Format, Args,
```
mrallen1: This macro construction means that lager's parse transform will not do the "whether to log this message" evaluation at the code site.  I am mild to the need for a macro based interface (I personally would not use it) but I understand people may want that kind of facility.

A better construction would be something like:
```erlang
-define(log(Level, Format, Args), lager:Level(Format, Args)).
-define(log_debug(Format, Args), ?log(debug, Format, Args)).
```

The `lager:Level` construction *will* get transformed by lager's parse transformation pass, and that should normally be the way that log messages get dispatched.

include/lager.hrl
```diff
@@ -20,6 +20,48 @@
...
+                           ?DEFAULT_TRUNCATION, Safety)).
+-define(log(Level, Metadata, Format, Args, Safety),
+        lager:dispatch_log(?DEFAULT_SINK, Level, ?METADATA(Metadata), Format, Args,
+                           ?DEFAULT_TRUNCATION, Safety)).
+
+-define(debug(Format, Args), ?log(debug, Format, Args, safe)).
```
mrallen1: Tiny bit concerned that lager's `?debug` and other log level macros might get caught up in a consuming project's macro definitions.

include/lager.hrl
tsloughter: I actually meant to include the 'whether to log this message' code in the macros body. I can add that.

And the point is to not have to have the parse transform pass, so replacing it with the `lager` call would defeat the purpose :)

include/lager.hrl
tsloughter: True, `?lager_debug` would be fine too.

tsloughter:
I'm not partial to parse transform or macro myself, both are fine to me. I figured it could be nice to have as an option, esp for the people who do take issue with parse transforms, hehe.

It is meant to be able to have all the same functionality as the transform, I think only `application`, which also isn't 100% always there even with the transform, is the only thing that can't be supported through the macro.

include/lager.hrl
mrallen1: Alright. You'll want to have a look at the `lager_transform` module injects then and use the macro to inline it.

mrallen1:
Alright. What do you think @Vagabond?

Vagabond:
It is something people have been asking for for a long time, I have no objection and would like to see it merged once it is ready.

tsloughter:
I'll get this added soon but I wanted to check on the use of the check in the macro. Is it simply to save a function call? Couldn't inlining be used for this just the same?

mrallen1:
I'm super jet lagged right now, but I don't understand what you're asking. Can you clarify please?

tsloughter:
The use of having the check currently done in `dispatch_log` moved to the call site (by either the parse transform or the macro) only saves a function call, right? So wouldn't telling the compiler to inline `dispatch_log` do the same?

mrallen1:
Without any more testing, it seems like you might be on to something, but I guess the only valid way to answer that question well is to benchmark doing it. :| 

(P.S. I am not asking you to sign up to do that testing unless you want to.)

tsloughter:
Nevermind, inlining in Erlang is only within the same module as the function. So can't inline something like `lager:dispatch_log` as far as i can tell.

Vagabond:
If only there was a way to modify the AST and inline a check so it could run very fast... :)

tsloughter:
I may have hit a deal breaker. 

The transform is able to generate unique variable names:

```
list_to_atom(Prefix ++ atom_to_list(get(module)) ++ integer_to_list(Line))
```

So as far as I can tell the macro can not get the same performance as the transform. People who are ideologically opposed to the transform may be fine with that, but I'd understand not wanting it to be an official part of lager.

But I did learn stuff about lager along the way :). I plan to see if performance improves by going back to using something like mochiglobal for the loglevel next.

Vagabond:
I would like to see the macro added, even if it is slower.

tsloughter:
Ok. Do you have an opinion on if it should be `?<level>` or `?lager_<level>`?

Vagabond:
I think we need to prefix it or we'd run into conflicts with existing macros (?debug for sure).

Vagabond:
Does this also include lager:md() metadata? I don't immediately see that it does...

tsloughter:
Ah, no. Adding.

mrallen1:
With respect to mochiglobal, have a look at https://github.com/mrallen1/fling :)

Vagabond:
I'd love to see fling as part of lager, actually. I originally removed mochiglobal support to try to reduce the dependancies lager had, but goldrush brought them back so there's not a lot of point sticking with ETS if fling/mochiglobal is better.

mrallen1:
I'm cool with that. For the record, my testing shows that at a certain number of items (in the 1000s) ETS outperforms the fling/mochiglobal model. For a very small number of elements, fling is (generally) faster IFF there are no new writes once the module is loaded into ERTS.

mrallen1:
Also the `?FUNCTION_NAME` macro breaks OTP 18 tests. :(

Vagabond:
Ah yes, maybe we can make that conditional on OTP version?

Also, if you want to make a PR integrating fling, I'd be very open to it.

mrallen1:
Opened an issue as a reminder/placeholder than this is a TODO item.

PR #427, opened 2017-12-19
Dont flush notify msgs at overload

uwiger:
We've been chasing weird timeout bugs in our test suites, presumably due to race conditions, but recently realized that lager was hitting the high-water mark and actually _deleting_ `app_started` events that our test suite depended on. This behavior violates the Principle of Least Surprise, to say the least.

In the interest of backward compatibility, I've made the behavior configurable, for `error_logger` and for each sink. I have, though, chosen the b/w incompatible default of _not_ flushing events. As long as it's properly documented and configurable, which should be the default can of course be discussed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,6 +356,24 @@ related processes crash, you can set a limit:
 
 It is probably best to keep this number small.
 
+### Event queue flushing
+
+When the high-water mark is exceeded, lager can be configured to flush all
+event notifications in the message queue. This can have unintended consequences
+for other handlers in the same event manager (in e.g. the `error_logger'), as
+events they rely on may be wrongly discarded. By default, this behavior is disabled,
+but can be controlled, for the `error_logger' via:
+
+```erlang
+{error_logger_flush_queue, true | false}
+```
+
+or for a specific sink, using the option:
+
+```erlang
+{flush_queue, true | false}
+```
+
 ### Sink Killer
 
 In some high volume situations, it may be preferable to drop all pending log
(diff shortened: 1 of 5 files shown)


Vagabond:
Do you mean the application started log message or the `dist_ac_app_started` from the distributed application controller?

I'd be *very* hesitant to change the default behaviour as, without the event flushing, we can't prevent the error_logger mailbox filling up and degrading or crashing the node.

uwiger:
The events will be flushed by `check_hwm()` anyway, only one at a time.

Again, the default behavior can be discussed, and there's something to be said for the default being whatever was the previous behavior. OTOH, nowadays, with separate sinks and the sink killer, I would at least like to question how obvious it is that flushing notify events directly from the message queue is really needed.

One could also imagine the option being to set a message queue length limit, where queue flushing kicks in. Since the hwm default is 50, one could hardly argue that the node is at risk already at 50 log messages within one second. Just firing up a swagger server gives 100 debug messages in one go, due to the hard-coded setting of 100 acceptors.

uwiger:
> Do you mean the application started log message or the dist_ac_app_started from the distributed application controller?

Yes, in our case the application started log message. Those are the most likely events to match on, I think, and there is no good alternative method.

I did consider pattern-matching on them in the discard loop, but still feel that flushing messages from a shared queue is a violation of design principles, and should at the very least be clearly documented and possible to turn off completely.

Vagabond:
I'm fine with adding support to disable the flushing, I just don't think we can or should change the default.

I'm not wild about draining the error_logger mailbox, for sure, but arguably error_logger is the one violating design principles by being completely async and crashing nodes when they're under load.

uwiger:
I changed the default to `true` (i.e. as before), and added a message queue flush threshold (`0` by default, which is also as before).

uwiger:
The tests pass on my machine. Is the CI failure something that happens occasionally? The failing test does look timing sensitive.

Vagabond:
Thanks, Ulf.

PR #429, opened 2017-12-31
Handle yet another stacktrace form

Vagabond:
Sometimes the stacktrace is a larger array of {M, F, A, Prop} tuples.
The handler code was expecting at most 2 tuples in the list. Relax this
requirement and add a test that provokes this edge-case.


mrallen1:
This all looks fine, but it looks like there's some busted unit tests in OTP 18 due to a recent merge.

PR #430, opened 2018-01-01
Take a stab at OTP 18 compatability with macro logging

Vagabond:


Vagabond:
This fails on OTP 18 for the timing issue, not for the macro issue.

PR #436, opened 2018-01-19
Make a few corrections to the recent mailbox flushing changes

Vagabond:
Turns out that previous shaper PR I merged was not quite baked, sorry.

* `should_flush` was case_clausing
* the default for error_logger queue flushing was not set to true
* the dropped message count was being incorrectly incremented

If this looks good, we should cut a patch release ASAP @mrallen1

### DIFF
--- a/src/error_logger_lager_h.erl
+++ b/src/error_logger_lager_h.erl
@@ -72,7 +72,7 @@ set_high_water(N) ->
 
 -spec init(any()) -> {ok, #state{}}.
 init([HighWaterMark, GlStrategy]) ->
-    Flush = application:get_env(lager, error_logger_flush_queue, false),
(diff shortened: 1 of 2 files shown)


mrallen1:
Do you want to do the release?

mrallen1:
Checklist is something like:

- [ ] Write changelog
- [ ] Update .app.src version
- [ ] Make new git tag
- [ ] Push new git tag
- [ ] Publish to hex

Vagabond:
I don't actually know how to use hex, maybe you could show me?

mrallen1:
Yeah! It's pretty easy - the docs are here: https://hex.pm/docs/rebar3_publish - main thing is getting the hex plugin installed as a "global plugin" which is part of that link.  Once you get your hex account I'll add you to lager owners so you can publish

PR #442, opened 2018-02-06
Pretty print lists of records at top level and field values

shino:
This PR makes `lager:pr/2,3` to pretty print lists of records.

The first commit rename the existing test suite to include test cases for list pretty printing into it.
The second commit makes change in `lager:pr`.


mrallen1:
Yay @shino!! 🎉

shino:
Thanks a lot, Mark!

PR #445, opened 2018-02-16
use "[{level,info}]" instead

getong:
In some projects using lager, without configuration, this error turns out:
```
WARNING: This is a deprecated console configuration. Please use "[{level,info}]" instead.
```
I think this patch will be work.


mrallen1:
Thanks for the patch. It was an oversight not to make it the default.

PR #446, opened 2018-02-21
Update Elixir section with information about parse_transform

surik:
According to the official changelog for elixir 1.5 parse_transform is
deprecated since this version. See
https://github.com/elixir-lang/elixir/blob/v1.5/CHANGELOG.md#4-deprecations


mrallen1:
Thanks for the update. I knew this was on the table but I didn't know it was included in 1.5

PR #449, opened 2018-03-27
Bugfix. flush_threshold not working.

IgorKarymov:
Fix incorrect option name validation.


mrallen1:
Thanks for the patch!

Issue #455, opened 2018-06-05
Initial Erlang/OTP 21 compatibility


michaelklishin:
## Background

This is a mostly placeholder issue for visibility. Team RabbitMQ is working towards [OTP 21 compatibility](https://github.com/rabbitmq/rabbitmq-server/issues/1616) and Lager is one dependency we have that is affected.

OTP 21 introduces a new logger and some assumption Lager has (e.g. about certain processes being available) are no longer true. This issue will be updated with details and proposed solutions/workarounds as @dumbbell and I dig them out.

The new logger is also raises the question about Lager's place in the post-OTP 21 ecosystem but let's keep that out of scope. At the very least there has to be an upgrade path for Lager users.

## Specific Issues

See OTP 21-rc.2 [release notes for details](http://erlang.org/download/otp_src_21.0-rc2.readme).

 * OTP-13295 deprecates `error_handler`. The claim of backwards compatibility only seems to be true to an extent. Lager depends on `error_handler` in several places, explicitly or implicitly: [one](https://github.com/erlang-lager/lager/blob/master/src/lager_app.erl#L229-L233), [two](https://github.com/erlang-lager/lager/blob/master/src/lager_handler_watcher.erl#L117), [three](https://github.com/erlang-lager/lager/blob/master/src/lager_app.erl#L256), and likely more.

Vagabond:
Yeah, I think the big compatibility issue is around whether error_logger is running. I believe this stuff is limited to the lager_app file (I'm not sure how your second link is tied to error_logger).

We've had some discussion around how to handle this on IRC, but I think simply checking for the error_logger registered process before trying to install/remove/restore error_logger handlers would go a long way to solving this particular issue.

Vagabond:
```Erlang/OTP 21 [RELEASE CANDIDATE 2] [erts-10.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]

Eshell V10.0  (abort with ^G)
1> application:ensure_all_started(lager).
{ok,[syntax_tools,compiler,goldrush,lager]}
2> lager:dispatch_log(lager_event, info, [{pid, self()}], "Hello from Erlang ~p", [erlang:system_info(otp_release)], 1024, safe).
ok
3> 08:40:30.455 [info] Hello from Erlang "21"```

This is the simplest possible change (I also found a fixed a bug). I will push up a branch shortly.

mrallen1:
[...]

PR #456, opened 2018-06-05
OTP 21 support

Vagabond:
To support OTP 21 we manually start error_logger if it is not present
(and lager is configured to install an error_logger handler) and we add
error_logger as a logger handler.

Longer term we should switch to installing our own logger handler, but
that is a larger task as we'd have to redo all the event parsing, if
that is even possible.

Related to #455

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 sudo: false
 language: erlang
 otp_release:
+  - 21.0-rc2
   - 20.0
   - 19.3
   - 18.3
--- a/rebar.config
+++ b/rebar.config
@@ -21,7 +21,7 @@
 
 {erl_opts, [
     {lager_extra_sinks, ['__lager_test_sink']},
-    {platform_define, "(19|20)", test_statem},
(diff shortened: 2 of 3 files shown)


michalwski:
@mrallen1, @Vagabond Do you think it's worth testing lager on travis with Erlang/OTP 21-rc1? It's possible already, I asked them to add it and they did :) We already try it with MongooseIM. All you need to do is add `21.0-rc1` to Erlang versions in `.travis.yml`

dumbbell:
Hi!

I confirm this fixes Lager used with RabbitMQ and Erlang 21.0-rc.2 for us.

michaelklishin:
Team RabbitMQ would greatly appreciate a new release with this patch ❤️

Vagabond:
I'm not really interested in testing RC1, but we can certainly add RC2 to the test matrix for a while.

I will add it to the travis file, let it rerun and if nobody objects we can cut a release today or tomorrow.

mrallen1:
👍

emqplus:
@Vagabond 👍

PR #457, opened 2018-06-05
Fix (dirty) for OTP 21

fenek:
This fixes `error_logger` integration for OTP 21. The current test suite passes (some errors logs may be seen on RC1 but RC2 works perfectly).

It is not a definite fix and should be replaced with proper `logger` integration in the future.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -21,7 +21,8 @@
 
 {erl_opts, [
     {lager_extra_sinks, ['__lager_test_sink']},
-    {platform_define, "(19|20)", test_statem},
+    {platform_define, "(21)", new_logging_api},
+    {platform_define, "(19|20|21)", test_statem},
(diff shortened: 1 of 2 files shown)


Vagabond:
This is very similar to #456 which I pushed this morning, do you think there's any advantages in this approach over the other one?

mrallen1:
First, thanks for the PR! Appreciate it!

I prefer #456, because it lays the ground work for a quote real unquote integration with the underlying logger API.  My 2 cents. What do you think @fenek?

fenek:
Hi!

Ooops, I did check for such PR today but it must have been just before Vagabond created it. :( Your approach seems better but I have to further verify both solutions because I still have some issues with `hut` library used by `exometer_core` (not sure if it's a problem with `hut` or `lager`).

mrallen1:
I'm going to close this, since we merged #456

Issue #458, opened 2018-06-17
Tracing settings are not restored after killer_hwm


russagit:

mrallen1:
I'm going to close this since we merged #459

PR #459, opened 2018-06-19
Fix reinstalling all handlers after killing a sink #458

russagit:


mrallen1:
I'll give this a closer look in a couple days but looks good to me on a quick review. Thanks!

mrallen1:
Thanks again for this PR.

PR #462, opened 2018-07-10
Add a wrapper to allow lager to install a sys trace function

Vagabond:
I did this a few weeks ago and it's been quite useful.


Vagabond:
Done.

PR #468, opened 2018-08-28
Fix markup - inline code

GrzegorzW:
It fixes this wring markup formatting:

screen shot 2018-08-28 at 21 48 59

The result:

screen shot 2018-08-28 at 21 52 32


mrallen1:
Thanks!

PR #469, opened 2018-08-29
Allow the console backend to redirect output to a remote node

Vagabond:
To support tracing to a console that is not the erlang node's *direct*
console introduce 2 additional options to the console backend:
`group_leader` and `id`. Group leader allows to use io:put_chars to send
the console output to a remote pid (eg. the group leader on a node
created by nodetool) and the ID allows multiple console backends to
exist simultaneously (for example {lager_console_backend, NodeName}).

Additionally, to support ephemeral console backends with associated
traces, if the console handler explicitly removes itself from the event
handler, remove any traces for that id.

To facilitate this, a new function has been added to the lager module:
`lager:clear_trace_by_destination/1`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -981,6 +981,32 @@ You can remove the trace when you're done by doing:
 lager:remove_trace(Pid).
 ```
 
+Console output to another group leader process
+----------------------------------------------
+
+If you want to send your console output to another group_leader (typically on
+another node) you can provide a `{group_leader, Pid}` argument to the console
+backend. This can be combined with another console config option, `id` and
+gen_event's `{Module, ID}` to allow remote tracing of a node to standard out via
+nodetool:
+
+```erlang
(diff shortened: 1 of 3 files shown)


Vagabond:
For an example of how to use it (in a function called from `nodetool rpc`):

```erlang
    GL = erlang:group_leader(),$
    Node = node(GL),$
    lager_app:start_handler(lager_event, {lager_console_backend, Node}, [{group_leader, GL}, {level, none}, {id, {lager_console_backend, Node}}]),$
    case lager:trace({lager_console_backend, Node}, Filter, Level) of
...
```

mrallen1:
This is a really cool feature - would you please add a section about it in the README and if possible a test for the configuration validation changes if not the feature directly.

mrallen1:
👍

mrallen1:
Thank you 😄

Issue #471, opened 2018-08-29
error_logger_lager_h exiting on gen_statem timeouts


PiedPieper:
Hi,

Should error_logger_lager_h be exiting on timeout events?

> handle_event(state_timeout, _Msg, StateName, State) ->
>    {stop, timeout};

I am exiting "gracefully" from a gen_statem timeout, but error_logger_lager_h is crashing:

> 21:02:04.591 [error] <0.584.0> Undefined:Undefined(Undefined):  Lager event handler error_logger_lager_h exited with reason {'EXIT',{{badmatch,[test_module,{state_timeout,timeout_event},[{data,[{"State",{init_delete,{state,undefined,443}}}]}],exit,timeout,[handle_event_function,state_enter],[{gen_statem,parse_event_result,8,[{file,"gen_statem.erl"},{line,1267}]},{gen_statem,loop_event,6,[{file,"gen_statem.erl"},{line,1015}]},{proc_lib,init_p_do_apply,3,[{file,...},...]}]]},...}}
> 21:02:04.610 [debug] <0.584.0> lager_handler_watcher:Undefined(119):  Lager installed handler error_logger_lager_h into error_logger

Vagabond:
Looks like you have found an edge case in the statem message parsing.

A good way to get a better error (to help us fix it) is to re-install the error_logger_tty_h event handler and provoke the crash again:

```erlang
gen_event:add_handler(error_logger, error_logger_tty_h, []).
```

PiedPieper:
=ERROR REPORT==== 29-Aug-2018::21:51:55 ===
** State machine test_module terminating
** Last event = {state_timeout,timeout_event}
** When server state  = [{data,
                             [{"State",
                               {init_auth,
                                   {state, undefined,
                                       443}}}]}]
** Reason for termination = exit:timeout
** Callback mode = [handle_event_function,state_enter]
** Stacktrace =
**  [{gen_statem,parse_event_result,8,[{file,"gen_statem.erl"},{line,1267}]},
     {gen_statem,loop_event,6,[{file,"gen_statem.erl"},{line,1015}]},
     {proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,247}]}]

=CRASH REPORT==== 29-Aug-2018::21:51:55 ===
  crasher:
    initial call: test_module:init/1
    pid: <0.662.0>
    registered_name: test_module
    exception exit: timeout
[...]

PR #473, opened 2018-08-30
Handle strange crash report from gen_statem

Vagabond:
This should be a fix for #471.

No direct test is provided because, as mentioned in the issue, I believe this is undefined behaviour.


Vagabond:
I'm going to merge this because it fixes the reported problem.

PR #474, opened 2018-09-03
rotation - default log file not deleted

### LINKED ISSUE #402: Lager does not create a new log file immediately after the old log file is rotated
After log rotation is not created new empty file.
This is required when used monitoring tools like zabbix to check file size.

Could you add ability to create new empty file after  log rotation.
P.S. As workaround may be used external log rotation.

```
[root@kazoo-app02 log]# ls -l
total 16920
-rw-r--r-- 1 kazoo daemon 2496086 Apr 12 07:02 console.log
-rw-r--r-- 1 kazoo daemon 8534767 Apr 11 23:59 console.log.1
-rw-r--r-- 1 kazoo daemon 5485187 Apr 10 23:59 console.log.3
-rw-r--r-- 1 kazoo daemon  675063 Apr 10 00:00 console.log.4
-rw-r--r-- 1 kazoo daemon       0 Apr 11 15:40 crash.log.1
-rw-r--r-- 1 kazoo daemon   69156 Apr  8 09:56 crash.log.4
-rw-r--r-- 1 kazoo daemon    2106 Apr 11 15:40 error.log.1
-rw-r--r-- 1 kazoo daemon   35375 Apr  8 09:56 error.log.4
```

shamis:
fixes #402

### DIFF
--- a/src/lager_rotator_default.erl
+++ b/src/lager_rotator_default.erl
@@ -121,7 +121,6 @@ rotate_file_fail_test() ->
     file:write_file(TestLog, "hello"),
     %% hose up the permissions
     os:cmd("chmod -R u-w " ++ TestDir),
-    rotate_logfile(TestLog, 10),
     ?assertMatch({error, _}, rotate_logfile(TestLog, 10)),
     ?assert(filelib:is_regular(TestLog)),
     %% fix the permissions
--- a/src/lager_rotator_default.erl
+++ b/src/lager_rotator_default.erl
@@ -75,7 +75,6 @@ rotate_logfile(File, 0) ->
     % opeing the file in write mode to clear the contents
     case file:open(File, [write]) of
         {ok, Fd} ->
-                file:open(File, [write]),
                 file:close(Fd);
         Error ->
                 Error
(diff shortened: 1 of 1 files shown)


Vagabond:
Oh, this is great! Very simple solution to the problem.

I will push something with a few changes shortly (and try to credit you on the commit).

Vagabond:
Take a look at https://github.com/erlang-lager/lager/commit/8d7eecda49d13b87a28efd5fb7f77185be263515 and see what you think.

shamis:
Thanks. Everything looks good. But do we need to return on error of `rotate_logfile(File, 1)` ?

shamis:
I think file descriptor has to be closed here https://github.com/erlang-lager/lager/commit/8d7eecda49d13b87a28efd5fb7f77185be263515#diff-1241b45549162d6b3da7af338c645bd3R78

Vagabond:
You're right about `rotate_logfile(File, 1)`, I realized that after I pushed.

The file descriptor will be closed by the garbage collector, we don't really have to worry about it.

I will push an update commit though.

Vagabond:
Ok, pushed an update. Thanks again for the contribution. Will close this issue.

mrallen1:
Oh, thanks so much @shamis. This issue has been open for a long long time!

shamis:
you are welcome. happy to help

PR #467, opened 2018-08-21
Prevent silent dropping of messages when hwm is exceeded

russagit:
I noticed that when flush_queue is switched off and hwm is exceeded messages are just silently dropped without any warning. I tried to fix this.

PR #475, opened 2018-09-06
Fix check_hwm

russagit:
During further testing found one more bug in check_hwm
This fix is an addition to https://github.com/erlang-lager/lager/pull/467

I have to add this one more fix. The logic appeared to be more complicated and I missed one more bug in check_hwm. Hope now it should be ok. The fix does not look beautiful, sorry.


getong:
@russagit 
I think your code is wrong, and I have made a pr to the same problem, and it works.
You can see it at #476 .

mrallen1:
👍

getong:
The mps should be reset to 0 , not 1.
And the test case is still failed.

PR #476, opened 2018-09-08
use the variable already exist

getong:


mrallen1:
Hi, appreciate the PR, but it looks like there's some problems running your tests in Travis. See for example: https://travis-ci.org/erlang-lager/lager/jobs/427249681

Can you take a look and address those please?

getong:
I will.

test/lager_test_function_transform.erl
```diff
@@ -122,10 +124,11 @@ transform_function_test_() ->
           ?assertEqual(1, lager_test_backend:count()),
           {_Level, _Time, _Message, Metadata} = lager_test_backend:pop(),
           Function = proplists:get_value(returns_dynamic_emit, Metadata),
-          ?assert(Function() < Function()),
-          ?assert(Function() < Function()),
-          ?assert(Function() < Function()),
-          ?assert(Function() < Function()),
+                ct:pal("Function:~p~n", [Function]),
```
mrallen1: Why did all of this get changed? To emit log messages in eunit, you need to use the `?debugFmt()` macro not ct:pal :)

getong: I am sorry, this is just my test code, I will remove it.

getong:
@mrallen1 
The pr is finally ok, you can check the code.

test/lager_test_backend.erl
```diff
@@ -1113,9 +1113,11 @@ error_logger_redirect_setup() ->
     application:load(lager),
     application:set_env(lager, error_logger_redirect, true),
     application:set_env(lager, handlers, [{?MODULE, info}]),
+    application:set_env(lager, suppress_supervisor_start_stop, false),
+    application:set_env(lager, suppress_application_start_stop, false),
     lager:start(),
     lager:log(error, self(), "flush flush"),
-    timer:sleep(100),
+    timer:sleep(1000),
```
mrallen1: It was a bad idea to introduce sleeps in these functions. There has to be a cleaner way to signal that we're ready to proceed with tests.  I think the way to do that is to have the `flush` handler emit a message that the flush is completed. But I guess that could be a new enhancement tackled down the road.

I'm thinking like 

```erlang
   Ref = make_ref(),
   gen_event:call(lager_event, ?MODULE, {flush, Ref, self()}),
   receive
       {Ref, flush_completed} -> lager_event
   after ?TEST_TIMEOUT ->
        erlang:error(test_timeout)
   end.
```

I don't know. Maybe that's a terrible idea but there's got to be something better than a sleep.

mrallen1:
Also it would be cool if you'd squash some of the commits you ended up abandoning. :)

mrallen1:
Not sure if you intended to squash commits but they're not squashed. :)

getong:
I recheck the commits, only the last five commits can be squashed into one commit. Or you mean to squash all commits into one commit?

getong:
You can also cherry pick some commits and squash them.

michaelklishin:
@getong some maintainers prefer all commits in a PR to be squashed into one.

mrallen1:
In this case, yes, it would be nice to have them squished into a commit.

getong:
ok, I will do it tonight.

getong:
done.

PR #480, opened 2018-09-25
Improve debug tracing support

Vagabond:
By an unhappy accident, OTP21 added functionality for 'named' trace
functions. This has a pattern match that matches the old trace function
(a 2 tuple where the second element is itself a 2 tuple):

https://github.com/erlang/otp/pull/1781

This patch changes to using a versioned record (so we can extend the
tracing later if needed) and adds 3 more options to tracing; count,
timeout and format string.

Additionally, tracing an OTP process via start_link arguments is now
documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -973,6 +973,18 @@ function in a target process by doing
 lager:install_trace(Pid, notice).
 ```
 
+You can also customize the tracing somewhat:
+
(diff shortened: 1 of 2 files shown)


mrallen1:
👍

PR #483, opened 2018-10-17
On non-Unix file systems, inode is zero.

zyuanfeng:
On windows, file:read_file_info/1, inode == 0.


zyuanfeng:
On windows, across the day, it don't work.

mrallen1:
Merged #493 which should fix this issue. Thanks for the PR.

PR #486, opened 2018-11-27
Remove check for erlang:is_map/1

Ledest:
Current lager requires Erlang/OTP >= 18 so no check required.


mrallen1:
Thanks for the patch.

PR #491, opened 2019-02-02
Try to use the new persistent term module instead of ETS in lager_config

Vagabond:


src/lager_config.erl
```diff
@@ -67,21 +60,76 @@ get(Key) ->
...
-    ets:insert(?TBL, {{Sink, Key}, Value});
+    insert({Sink, Key}, Value);
 set(Key, Value) ->
     set({?DEFAULT_SINK, Key}, Value).
+
+%% check if we can use persistent_term for config
+%% persistent term was added in OTP 21.2 but we can't
+%% check minor versions with macros so we're stuck waiting
+%% for OTP 22
+-if(?OTP_RELEASE < 22).
```
mrallen1: This preprocessor syntax sadly only works in >= 21, the compiler for OTP 20 and 19 barf on it.

src/lager_config.erl
```diff
@@ -67,21 +60,76 @@ get(Key) ->
...
+    _ = try ets:new(?TBL, [named_table, public, set, {keypos, 1}, {read_concurrency, true}]) of
+            _Result ->
+                ok
+        catch
+            error:badarg ->
+                ?INT_LOG(warning, "Table ~p already exists", [?TBL])
+        end.
+
+insert(Key, Value) ->
+    ets:insert(?TBL, {{Sink, Key}, Value}).
```
mrallen1: Think this is a copy/paste error because there's no Sink defined in the insert function head.

Vagabond:
Yeah, turns out this needs more work, bleh.

Vagabond:
I think I've addressed the problems with this...

Vagabond:
Thanks!

geogriff-signal:
are you sure this was a good idea? from the documentation of `persistent_term`:
> When a (complex) term is deleted (using erase/1) or replaced by another (using put/2), a global garbage collection is initiated. It works like this:
>
> All processes in the system will be scheduled to run a scan of their heaps for the term that has been deleted. While such scan is relatively light-weight, if there are many processes, the system can become less responsive until all process have scanned their heaps.

... and:
> One way to minimize the effects on responsiveness could be to minimize the number of processes on the node before updating or deleting a persistent term. It would also be wise to avoid updating terms when the system is at peak load.

It seems like setting persistent terms after startup is in general not suitable for systems running a large number of processes, which lager aught to support.

Vagabond:
I don't believe we qualify as a '(complex)' term:

```
Deletion of atoms and other terms that fit in one machine word is specially optimized to avoid doing a global GC
```

Vagabond:
You might be right that we are accidentally storing a 'complex' term, now that I review what we're storing. I will have to look at it in more detail later.

Issue #500, opened 2019-05-20
Resolve new dialyzer errors


vikger:
Hi,

The following errors were produced when running dialyzer on the latest lager on master branch

```
$ make dialyzer
test -d deps || ./rebar get-deps
./rebar compile
==> goldrush (compile)
==> lager (compile)
Compiled src/lager_util.erl
Compiled src/lager_app.erl
Compiled src/lager_handler_watcher.erl
Compiled src/lager.erl
Compiled src/error_logger_lager_h.erl
Compiled src/lager_rotator_default.erl
Compiled src/lager_file_backend.erl
  Checking whether the PLT /Users/viktorgergely/.combo_dialyzer_plt is up-to-date... yes
done (passed successfully)
 yes
done (passed successfully)
  Checking whether the PLT .local_dialyzer_plt is up-to-date... yes
done (passed successfully)
 yes
done (passed successfully)
==> lager (dialyzer)
error_logger_lager_h.erl:367: The variable Invalid can never match since previous clauses completely covered the type 'logged' | 'no_log'
lager.erl:694: Expression produces a value of type 'ok' | {'error','lager_not_running'}, but this value is unmatched
lager_app.erl:170: Expression produces a value of type 'ok' | {'error',_}, but this value is unmatched
lager_file_backend.erl:198: Expression produces a value of type 'ok' | #state{name::string(),level::{'mask',integer()},fd::'undefined' | pid() | {'file_descriptor',atom(),_},inode::'undefined' | integer(),flap::boolean(),size::integer(),date::'undefined' | string(),count::integer(),rotator::atom(),shaper::#lager_shaper{hwm::'undefined' | pos_integer(),mps::non_neg_integer(),lasttime::{_,_,_},dropped::non_neg_integer(),flush_queue::boolean(),flush_threshold::integer(),timer::reference(),filter::fun()},formatter::atom(),sync_on::{'mask',integer()},check_interval::non_neg_integer(),sync_interval::non_neg_integer(),sync_size::non_neg_integer(),last_check::{non_neg_integer(),non_neg_integer(),non_neg_integer()}}, but this value is unmatched
lager_handler_watcher.erl:78: Expression produces a value of type {'error',_} | {'ok',timer:tref()}, but this value is unmatched
lager_handler_watcher.erl:101: Expression produces a value of type {'error',_} | {'ok',timer:tref()}, but this value is unmatched
[...]

PR #501, opened 2019-05-21
Fix dialyzer errors

vikger:
This pull request has been created to solve issue #500. Please review.

Change details:
- Unnecessary case branch removed in error_logger_lager_h.erl:367 (?LOGFMT returns logged | no_log).
- Added pattern "_ =" to cases where return value of the expression is not used.

Regards,
Viktor


mrallen1:
All green. Thanks.

PR #514, opened 2019-08-13
Metadata whitelist

macpie:
Allow to create a metadata whitelist in `sys.config` so that you do not have to create a custom formatter.

```
[{lager, [
    {metadata_whitelist, [request_id]}
]}]
```

It will print them (using the default formatter):

```... [error]  request_id=123 Message```


Vagabond:
Although I wish we could do this per-handler, it seems like there's too much legacy cruft in the way to get that done just now. This seems like a modest improvement at least.

PR #515, opened 2019-08-17
Replace string:find/2 for compatibility with Erlang/OTP 19

Ledest:
Erlang/OTP 19 does not have string:find so replace it to compatable methods.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,7 @@ otp_release:
   - 22.0
   - 21.1
   - 20.3
-
+matrix:
+  include:
+  - otp_release: 19.3
+    dist: trusty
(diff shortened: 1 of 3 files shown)


Vagabond:
We only support the last 3 major releases of Erlang/OTP and OTP 19 is now beyond that window.

Ledest:
> OTP 19 is now beyond that window.

So README is lying?

macintux:
README includes the key phrase “as of December 2018”, so it’s outdated but explicitly written to be clear about its limitations.

22.0 was released in May.

Vagabond:
I would also suggest you not jump to the conclusion we're not trying to deliberately mislead and instead consider the possibility that we simply have not updated the README (which I will do now).

If you want OTP 19 support you can certainly apply your patch locally, or use one of the older releases from before we dropped support for it. Unfortunately we don't have a lot of support bandwidth these days, so we have to pick our battles on what we can support.

Ledest:
I apologize, I was just misled by an not updated README.

PR #516, opened 2019-08-20
Allow trace filter to silence log output

uwiger:
Allow a trace pattern `{Dest, Filter, silence}`, which suppresses matching log messages for the given destination.

This is mainly intended for when logs are cluttered with output not relevant to debugging efforts.

### DIFF
--- a/src/lager_util.erl
+++ b/src/lager_util.erl
@@ -77,6 +77,8 @@ num_to_level(?EMERGENCY) -> emergency;
 num_to_level(?LOG_NONE)  -> none.
 
 -spec config_to_mask(atom()|string()) -> {'mask', integer()}.
+config_to_mask(silence) ->
+    silence;
 config_to_mask(Conf) ->
     Levels = config_to_levels(Conf),
     {mask, lists:foldl(fun(Level, Acc) ->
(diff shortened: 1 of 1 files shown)


uwiger:
WIP, no tests written yet.

uwiger:
A note: I know bits are missing - not least documentation. I'd like some feedback re. the general idea first. I know that this does violate the original idea of tracing, in a sense.

mrallen1:
I like the general idea a lot. How is `silence` different from `none`?

mrallen1:
Hey @uwiger any further updates on this?

uwiger:
Sorry for the very long delay. :)

This became relevant again as we tried to switch over to the latest upstream lager.

To answer the question on how this differs from `none`, I believe that `none` effectively means `0 bor LevelMask`, which doesn't actively turn on tracing for a given filter, but also didn't shut it up if the existing mask specified a certain level.

Our issue was that a certain module in a 3rd party app was reporting too much at 'debug' level, and we wanted to shut it up.

mrallen1:
Some tests and docs would be great - then we can get this merged in - thank you

PR #524, opened 2020-03-01
Enable lager to use the OTP logger's event pipeline

Vagabond:
We should really get this finished so let's try to figure out what's missing.

### DIFF
--- a/src/lager.erl
+++ b/src/lager.erl
@@ -36,7 +36,8 @@
         get_loglevel/1, get_loglevel/2, set_loglevel/2, set_loglevel/3, set_loglevel/4, get_loglevels/1,
         update_loglevel_config/1, posix_error/1, set_loghwm/2, set_loghwm/3, set_loghwm/4,
         safe_format/3, safe_format_chop/3, unsafe_format/2, dispatch_log/5, dispatch_log/7, dispatch_log/9,
-        do_log/9, do_log/10, do_log_unsafe/10, pr/2, pr/3, pr_stacktrace/1, pr_stacktrace/2]).
+        do_log/9, do_log/10, do_log_unsafe/10, pr/2, pr/3, pr_stacktrace/1, pr_stacktrace/2,
+        configure_logger/0]).
 
 -type log_level() :: none | debug | info | notice | warning | error | critical | alert | emergency.
 -type log_level_number() :: 0..7.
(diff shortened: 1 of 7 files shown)


Vagabond:
Some things we need to do:

- [ ] Documentation
- [ ] A way to export the current lager config as a logger config
- [ ] An optional way to, in lager:start(), turn the lager config into a logger config programatically
- [ ] Figure out which of the more advanced lager features can still be supported in logger

src/lager.erl
```diff
@@ -689,6 +690,64 @@ rotate_handler(Handler) ->
...
+generate_logger_handlers([{lager_file_backend, Config}|Tail], {CurrentLevel, Acc}) ->
+    Level = proplists:get_value(level, Config, info),
+    File = proplists:get_value(file, Config),
+    LogRoot = application:get_env(lager, log_root, ""),
+    Size = proplists:get_value(size, Config),
+    Count = proplists:get_value(count, Config),
+    Formatter = proplists:get_value(formatter, Config, lager_default_formatter),
+    FormatterConfig = proplists:get_value(formatter_config, Config, []),
+    %% XXX I don't think disk_log is suitable as it is, we should provide a logger compatible version of
+    %% lager's file_backend and try to patch disk_log upstream
```
mrallen1: Agree

src/lager_sup.erl
```diff
@@ -32,24 +32,29 @@ start_link() ->
...
-    {ok, {{one_for_one, 10, 60},
-          Children ++ CrashLog
-         }}.
+    case application:get_env(lager, lager_use_logger, false) of
+        true ->
+            {ok, {{one_for_one, 10, 60}, []}};
+        false ->
+            %% set up the config, is safe even during relups
+            lager_config:new(),
+            %% TODO:
```
mrallen1: Nitpick: probably should delete or update this comment

src/lager_logger_formatter.erl
```diff
@@ -0,0 +1,119 @@
...
+format(#{level := Level, msg := {string, String}, meta := Metadata}, Config) ->
+    do_format(Level, String, Metadata, Config);
+format(#{level := Level, msg := {FmtStr, FmtArgs}, meta := Metadata}, Config) ->
+    Msg = lager_format:format(FmtStr, FmtArgs, maps:get(max_size, Config, 1024)),
+    do_format(Level, Msg, Metadata, Config).
+
+do_format(Level, Msg, Metadata, Config) ->
+    FormatModule = maps:get(formatter, Config, lager_default_formatter),
+    Timestamp = maps:get(time, Metadata),
+    MegaSecs = Timestamp div 1000000000000,
```
mrallen1: Make these numbers macros to improve readability and maintenance sanity, please.

jadeallenx (commented):
Looks good in general. I haven't investigated the failing tests yet but I suspect they're just due to the WIP nature of these changes.

src/lager_logger_formatter.erl
```diff
@@ -0,0 +1,119 @@
...
+report_cb(#{label := {application_controller, exit}, report := Report}) ->
+    {exited, Reason} = lists:keyfind(exited, 1, Report),
+    case application:get_env(lager, suppress_application_start_stop) of
+        {ok, true} when Reason == stopped ->
+            {"", []};
+        _ ->
+            {application, Name} = lists:keyfind(application, 1, Report),
+            Formatted = error_logger_lager_h:format_reason(Reason),
+            {"Application ~w exited with reason: ~s", [Name, Formatted]}
+    end.
```
hauleth: This will fail on custom reports, ex.:

```erlang
?LOG_INFO(#{http_reponse => 200, request_duration => Duration})
```

src/lager.erl
hauleth: Maybe use `logger_std_h` with `file` option set. This should result in more suitable solution.

src/lager.erl
```diff
@@ -689,6 +690,64 @@ rotate_handler(Handler) ->
 rotate_handler(Handler, Sink) ->
     gen_event:call(Sink, Handler, rotate, ?ROTATE_TIMEOUT).
 
+generate_logger_config() ->
+    Handlers = application:get_env(lager, handlers, lager_app:default_handlers()),
+    {Level, NewHandlers} = generate_logger_handlers(Handlers, {notice, []}),
```
hauleth: Maybe instead of "translating" Lager handlers (which will not support custom handlers anyway) to logger handlers, add new handler that will forward all messages from the logger to lager's `gen_event` as a compatibility layer.

src/lager.erl
```diff
@@ -689,6 +690,65 @@ rotate_handler(Handler) ->
...
+    Acc;
+generate_logger_handlers([{lager_console_backend, Config}|Tail], {CurrentLevel, Acc}) ->
+    Level = proplists:get_value(level, Config, info),
+    Formatter = proplists:get_value(formatter, Config, lager_default_formatter),
+    FormatterConfig = proplists:get_value(formatter_config, Config, []),
+    Handler = {handler, console, logger_std_h, #{level => Level, formatter =>
+                                                 {lager_logger_formatter, #{report_cb => fun lager_logger_formatter:report_cb/1,
+                                                                            formatter => Formatter,
+                                                                            formatter_config => FormatterConfig}}}},
+    NewLevel = case lager_util:level_to_num(Level) > lager_util:level_to_num(CurrentLevel) of
```
hauleth: Instead of that you can use [`logger:compare_levels/2`](http://erlang.org/doc/man/logger.html#compare_levels-2).

PR #529, opened 2020-08-25
add a pterm formatter

evanmcc:
This PR adds a formatting command `{pterm, key, DefaultVal}`.

This is meant for adding global things to the formatting string, which can be set once, on startup, rather than repeatedly stuffed into the pid-specific metadata.  My particular use case is the OS pid, so it's easier to tell when, in the logs, a particular process restarted.


mrallen1:
The only hesitation I have is that persistent_term was introduced in OTP 21.2, so the test should skip if OTP is less than that. But I can fix that up

mrallen1:
closing in favor of #530

PR #532, opened 2020-10-29
Fix some tests

paulo-ferraz-oliveira:
I get odd crashes in OTP 22+23, related to the following:

```erlang
-        TestBody("gen_statem crash", crash_statem, crash, [], "gen_statem crash_statem in state state1 terminated with reason: no function clause matching crash_statem:handle"),
-        TestBody("gen_statem stop", crash_statem, stop, [explode], "gen_statem crash_statem in state state1 terminated with reason: explode"),
-        TestBody("gen_statem timeout", crash_statem, timeout, [], "gen_statem crash_statem in state state1 terminated with reason: timeout")
+        TestBody("gen_statem crash", crash_statem, crash, [], "gen_statem crash_statem in state state1 terminated with reason: {function_clause,handle_event_function}"),
+        TestBody("gen_statem stop", crash_statem, stop, [explode], "gen_statem crash_statem in state state1 terminated with reason: {explod"),
+        TestBody("gen_statem timeout", crash_statem, timeout, [], "gen_statem crash_statem in state state1 terminated with reason: {timeou")
```

Didn't investigate deeper, since I wanted to push this to understand if it's OK. (I wonder if the errors were there already?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,8 @@ otp_release:
   - 23.0
   - 22.3
   - 21.3
+script:
+  - rebar3 --version
+  - erl -version
+  - rebar3 dialyzer
+  - rebar3 eunit
--- a/src/lager_util.erl
+++ b/src/lager_util.erl
@@ -32,7 +32,7 @@
 ]).
 
 -ifdef(TEST).
--export([create_test_dir/0, get_test_dir/0, delete_test_dir/0,
+-export([create_test_dir/0, get_test_dir/0, delete_test_dir/0, delete_test_dir/1,
          set_dir_permissions/2,
          safe_application_load/1,
          safe_write_file/2]).


src/error_logger_lager_h.erl
```diff
@@ -632,6 +632,7 @@ no_silent_hwm_drops_test_() ->
         [
             fun() ->
                 error_logger:tty(false),
+                ok = application:load(goldrush),
```
paulo-ferraz-oliveira, line 635: For OTP 19, at least.

.travis.yml
```diff
@@ -4,3 +4,10 @@ otp_release:
   - 23.0
   - 22.3
   - 21.3
+  - 20.3
```
paulo-ferraz-oliveira, line 7: This file is apparently not used, anymore. Are we testing for OTP 19, in any case?

mrallen1:
We don't test on OTP 19 any more, as it's outside of our support window - however, thanks for the patch to fix up tests on 22 and 23. Much appreciated!

paulo-ferraz-oliveira:
> it's outside of our support window

Nice. Should `rebar.config` signal it, though? It would mean it wouldn't compile (under `rebar3` at least) below a minimum OTP version.

mrallen1:
There are a lot of lager dependencies which need to run on older OTP releases so putting it in rebar.config just feels like an annoying and artificial restriction for those builds to me.

paulo-ferraz-oliveira:
@mrallen1, do you want me to remove 19 and 20 from `.travis.yml`? (if the tests are passing it's not very bad to have it, is it?)

Also, do you think it would be worth understanding where the issues in the description come from? It's very odd.

e.g.
```
expects: with reason: explode
   gets: with reason: {explod
```

mrallen1:
Yeah 19 and 20 are both out of policy now. Just haven't removed them from travis yet.

paulo-ferraz-oliveira:
> Yeah 19 and 20 are both out of policy now. Just haven't removed them from travis yet.

Well, you had, but I added them back, since I assumed your support was 19+. I'll revert this change.

paulo-ferraz-oliveira:
I'll leave it up to you to decide to merge as-is or consider if it's worth running around the strange issues described in the first message.

mrallen1:
The real issue that the OTP team keeps changing the strings they emit for crashes from `gen_statem` - has changed every single release from the one where they were introduced. Probably will change in OTP 24 too :stuck_out_tongue_closed_eyes: We _should_ fix them but no one has been especially motivated to do so :(

paulo-ferraz-oliveira:
Understood, but the detail is what's intriguing. It's `{explod` instead of `explode`.
Any way, I could fix it, but if you say it keeps changing it's probably not worth it, since we'd have to keep coming up with exceptions for this.

paulo-ferraz-oliveira:
Is there anything missing to move forward?

mrallen1:
Trying to decide if we should delete the failing tests (or comment them out) or just buckle down and fix them...

paulo-ferraz-oliveira:
> Trying to decide if we should delete the failing tests (or comment them out) or just buckle down and fix them...

I could fix them but I'd need to understand your expectations. The easiest way to do it be to just compare the actual output with the expected output and fix it there (but we'd have to target different OTP versions, I supposed).

Did you have the change to look at the results of got v. expected (because that's what's really odd to me).

mrallen1:
Yeah it seems like https://github.com/erlang-lager/lager/blob/master/src/error_logger_lager_h.erl#L221 isn't handling the `Reason` we're trying to extract there - so it's using the naked tuple instead of a more human friendly string

paulo-ferraz-oliveira:
What would be the expected fix, then? Change the function or change the tests?

mrallen1:
The intention of that code is to take scary Erlang error tuples and provide a less scary, more humane interpretation of the error tuple, so it should be to take whatever error tuple we're getting and return a string like the one expected in the test. (Both the code *and* the test may need to change)

mrallen1:
Looks like some commits are missing here? But thanks for getting back to this

paulo-ferraz-oliveira:
> Looks like some commits are missing here? But thanks for getting back to this

Why do you feel there's commits missing? The ones that were there (that I force-pushed took out) related to OTP-19, which you said you didn't support anymore 😄 As-is, there's only 6 test cases failing for OTP 22 and OTP 23. I looked into the code but am still having a hard time finding where the issues come from. Will probably try again later (or eventually close the PR if I'm unable to find time and/or the source of the issues).

paulo-ferraz-oliveira:
Hi. I don't think I'll have time to tackle this further, any time soon, so I'm closing it, since I don't want this to be a "pending action item". Thanks for your feedback, though.

Issue #304, opened 2015-12-14
Does trace to file work properly? [JIRA: RIAK-2326]


malotte:
When adding traces using lager:trace_file/3 it seems like the log_root is added twice.
I. e. when using log_root = "log/lager" the file ends up at log/lager/log/lager.
Furthermore, no trace output is written to the file as far as I can see.

Add trace:
(hoc33@venus)2> lager:trace_file("debug.log", [{module, nmea_2000_log}], debug). 
{ok,{{lager_file_backend,"log/lager/debug.log"},
     [{module,nmea_2000_log}],
     debug}}
(hoc33@venus)3> lager:status().                                                 Lager status:
File log/lager/debug.log (lager_event) at level none
File log/lager/http.log (lager_event) at level none
Console (lager_event) at level info
File error.log (lager_event) at level warning
File console.log (lager_event) at level info
Active Traces:
Tracing messages matching {all,[{module,'=',nmea_2000_log}]} (sink lager_event) at level debug to {lager_file_backend,
                                                                                                   "log/lager/debug.log"}
Tracing messages matching {all,[{module,'=',seaz_ws},
                                {function,'=',handle_http_request}]} (sink lager_event) at level debug to {lager_file_backend,
                                                                                                           "log/lager/http.log"}
Tracing Reductions:
{any,[{all,[{function,'=',handle_http_request},{module,'=',seaz_ws}]},
         {module,'=',nmea_2000_log}]}
Tracing Statistics:
  input: 0
 output: 0
 filter: 0
ok

Log:
hoc33@venus)4> lager:log(debug, [{module, nmea_2000_log}], "test").
ok
(hoc33@venus)5> lager:status().                                     
Lager status:
File log/lager/debug.log (lager_event) at level none
File log/lager/http.log (lager_event) at level none
Console (lager_event) at level info
File error.log (lager_event) at level warning
File console.log (lager_event) at level info
Active Traces:
Tracing messages matching {all,[{module,'=',nmea_2000_log}]} (sink lager_event) at level debug to {lager_file_backend,
                                                                                                   "log/lager/debug.log"}
Tracing messages matching {all,[{module,'=',seaz_ws},
                                {function,'=',handle_http_request}]} (sink lager_event) at level debug to {lager_file_backend,
                                                                                                           "log/lager/http.log"}
Tracing Reductions:
{any,[{all,[{function,'=',handle_http_request},{module,'=',seaz_ws}]},
      {module,'=',nmea_2000_log}]}
[...]

PR #317, opened 2016-01-26
Do not add logroot in expand_path sometimes

mrallen1:
If expand_path/1 already contains LogRoot in its RelPath, do not add it again. Addresses the problem in #304 (RIAK-1367) (RIAK-2326)

PR #535, opened 2020-12-31
Ensure relative log file not change when current path changed

### LINKED ISSUE #534: when change pwd, RelPath will also change, and create new log file in new dir
with env
```erlang
{lager, [
        {log_root, "./log"},
        {crash_log, "crash.log"}
    ]
    },
```
after call `cd("log").` 
crash log will make new file in ./log/log/crash.log not continue write in ./log/crash.log
seems because when ensure log file，use rel file name，and rel file changed because of pwd changed
https://github.com/erlang-lager/lager/blob/91575d4479c897451edf138c10fa86c3125ec52e/src/lager_util.erl#L601-L604
then will create new log file
https://github.com/erlang-lager/lager/blob/91575d4479c897451edf138c10fa86c3125ec52e/src/lager_rotator_default.erl#L44-L47

same behavior in lager_file_backend, not sure this is mean to.
if not seems just change laget_util:expand_path from rel path to abs will resolve this，if this is fine，i could make a pr for this.

if mean to, maybe add a env key like `use_abs_path` or `abs_log_root` let config control is a way too

yidayoung:
Fix #534 
- Change lager_util:expand_path function will return abspath not relative
- Fix testcase of expand_path
- Delete function filename:absname call in lager_util:find_file，because of lager_util:expand_path will return abspath

### DIFF
--- a/src/lager_util.erl
+++ b/src/lager_util.erl
@@ -460,19 +460,21 @@ i3l(I)              -> integer_to_list(I).
 
 %% When log_root option is provided, get the real path to a file
 expand_path(RelPath) ->
-    case application:get_env(lager, log_root) of
-        {ok, LogRoot} when is_list(LogRoot) -> % Join relative path
-            %% check if the given RelPath contains LogRoot, if so, do not add
-            %% it again; see gh #304
-            case string:str(filename:dirname(RelPath), LogRoot) of
-                X when X > 0 ->
-                    RelPath;
-                _Zero ->
-                    filename:join(LogRoot, RelPath)
-            end;
-        undefined -> % No log_root given, keep relative path
-            RelPath
-    end.
+    RelPath2 = case application:get_env(lager, log_root) of
+                   {ok, LogRoot} when is_list(LogRoot) -> % Join relative path
+                       %% check if the given RelPath contains LogRoot, if so, do not add
+                       %% it again; see gh #304
+                       case string:str(filename:dirname(RelPath), LogRoot) of
+                           X when X > 0 ->
+                               RelPath;
+                           _Zero ->
+                               filename:join(LogRoot, RelPath)
+                       end;
+                   undefined -> % No log_root given, keep relative path
+                       RelPath
+               end,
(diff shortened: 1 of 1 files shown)


src/lager_util.erl
```diff
@@ -460,19 +460,21 @@ i3l(I)              -> integer_to_list(I).
...
-            end;
-        undefined -> % No log_root given, keep relative path
-            RelPath
-    end.
+    RelPath2 = case application:get_env(lager, log_root) of
+                   {ok, LogRoot} when is_list(LogRoot) -> % Join relative path
+                       %% check if the given RelPath contains LogRoot, if so, do not add
+                       %% it again; see gh #304
+                       case string:str(filename:dirname(RelPath), LogRoot) of
+                           X when X > 0 ->
```
paulo-ferraz-oliveira, line 468: This seems to work badly for e.g. `console.log`  and `log`, since root is part of the former, in terms of matching.

yidayoung: Yes, `string:str` is not so precise, this change commited in #317 to fix #304.
this pr change nothing with it.  

------

maybe should make a new pr or issue to specially discuss this,
 i saw https://github.com/erlang-lager/lager/issues/304#issuecomment-164807302 said should trans path to abspath. not not do this finally, maybe there has some other to avoid

paulo-ferraz-oliveira: Yeah, sorry, my bad. I didn't mean to imply your pull request changed it in any way. It was simply the most recent one I found "dealing" with this.

I think ```https://github.com/erlang-lager/lager/commit/b9ce226bff4b47f93a4aeac6bbd34bcf78412a8d``` might to the trick: it's part of another pull request I'm currently preparing. Maybe @mrallen1 can help out, here.

paulo-ferraz-oliveira: The origin might disappear so I'm pasting the code here `filename:dirname(RelPath) =:= filename:basename(LogRoot)`: it basically compares the first folder.

PR #540, opened 2021-02-09
Prevent "wrong" log location

paulo-ferraz-oliveira:
Currently, if `log_root` is e.g. `log` and `console.log` is not specified with a "folder", `log_root`'s value is not taken into account (give the nature of `src`).

This proposed approach does not, however, allow for e.g. `log/console.log` to be saved in `logs/log/console.log` (in case `log_root` is defined), but I believe that's not the common use; otherwise we might add a new option (always use `log_root`) that would allow for this mis-behaviour to occur.

---

Of course, to be perfectly strict, and do away with these workarounds, we'd always apply `log_root` as a prefix, independent of what is chosen as file name, since that's the semantics of `log_root`, it seems.

### DIFF
--- a/src/lager_util.erl
+++ b/src/lager_util.erl
@@ -464,11 +464,11 @@ expand_path(RelPath) ->
                    {ok, LogRoot} when is_list(LogRoot) -> % Join relative path
                        %% check if the given RelPath contains LogRoot, if so, do not add
                        %% it again; see gh #304
-                       case string:str(filename:dirname(RelPath), LogRoot) of
-                           X when X > 0 ->
-                               RelPath;
-                           _Zero ->
-                               filename:join(LogRoot, RelPath)
+                       case filename:dirname(RelPath) of
+                           "." ->
+                               filename:join(LogRoot, RelPath);
+                           false ->
+                               RelPath
                        end;
                    undefined -> % No log_root given, keep relative path
                        RelPath


paulo-ferraz-oliveira:
Bump.

mrallen1:
I'm generally 👍 on this but I agree it's sort of a trivial config issue.

paulo-ferraz-oliveira:
> This seems like a silly workaround for a configuration issue. I don't have a strong objection to this, although I agree that log_root should be used as a unilateral prefix if the logfile's name is not an absolute path.

@Vagabond: do you prefer it if I change the pull request in order to have log root be prepended to the file names in all cases **except** if the file's path starts with '/' (I'm not sure if there's a better way to determine it as _absolute_).

> I'm generally 👍 on this but I agree it's sort of a trivial config issue.

@mrallen1: I agree it's trivial, but it's also misleading (as-was implemented, at least, to me). 😄

With
```
{log_root, "log"},
{crash_log, "crash.log"}
```

I'd expect the log to be kept in `log/crash.log`, but it isn't, because of the way the `log` text is being detected (it goes to `crash.log`, at the root).

With
```
{log_root, "logs"},
{crash_log, "blogs.log"}
```

it's again the same unexpected behaviour (since `blogs` contains `logs`).

---

My setup

```
{log_root, "log"},
{crash_log, "crash.log"},
{console_log, "console.txt"} % this is not the name of the option
```

means I'll have info at
```
crash.log
log/console.txt
```

(I'd very much like to `.gitignore` a single folder, but in this case can't 😛)

paulo-ferraz-oliveira:
The solution I implemented allows

```
{log_root, "log"},
{crash_log, "crash.log"}
```

to keep the file at `log/crash.log`.

It'll also mean that 
```
{log_root, "log"},
{crash_log, "log/crash.log"}
```

will point to the same file (as I argued that I don't think - though I have no hard data - this is a usual case).

Vagabond:
Your patch is unquestionably better and is probably less disruptive than removing the check altogether.

mrallen1:
Thank you

michaelklishin:
This affected RabbitMQ node startup. The culprit seems to be how we set `lager.crash_log` to `"log/crash.log"`. First is a relevant trace snippet:

```
{{case_clause,"log"},
                                  [{lager_util,expand_path,1,
                                    [{file,"src/lager_util.erl"},{line,467}]},
                                   {lager_crash_log,init,1,
                                    [{file,"src/lager_crash_log.erl"},
                                     {line,76}]},
                                   {gen_server,init_it,2,
                                    [{file,"gen_server.erl"},{line,417}]},
                                   {gen_server,init_it,6,
                                    [{file,"gen_server.erl"},{line,385}]},
                                   {proc_lib,init_p_do_apply,3,
                                    [{file,"proc_lib.erl"},{line,226}]}]}}}}},
                             [{lager_app,start,2,
                               [{file,"src/lager_app.erl"},{line,227}]}
```

The `lager` app environment looks like this

```
[{log_root,"/var/folders/gp/53t98z011678vk9rkcb_s6ph0000gn/T/rabbitmq-test-instances/rabbit@hostname/log"},
                {error_logger_hwm,50},
                {crash_log,"log/crash.log"},
                {crash_log_count,5},
                {async_threshold_window,5},
                {crash_log_msg_size,65536},
                {crash_log_date,"$D0"},
                {colors,[{debug,"\e[0;34m"},
                         {info,"\e[1;37m"},
                         {notice,"\e[1;36m"},
                         {warning,"\e[1;33m"},
                         {error,"\e[1;31m"},
                         {critical,"\e[1;35m"},
                         {alert,"\e[1;44m"},
                         {emergency,"\e[1;41m"}]},
                {colored,true},
                {async_threshold,20},
                {handlers,[]},
                {crash_log_size,10485760},
                {error_logger_redirect,true},
                {crash_log_rotator,lager_rotator_default}]
```

I'm investigating what would be the best way to work around this change but I don't think we were doing anything controversial with our settings :/

michaelklishin:
After changing `crash_log` to just a filename and eradicating some of our own uses of `lager_util:expand_path/1` the issue goes away. This is a breaking change for any `lager_util:expand_path/1` user, though, and [release notes](https://github.com/erlang-lager/lager/releases/tag/3.9.0) for `3.9.0` are blank :/

paulo-ferraz-oliveira:
@michaelklishin: it can be considered a breaking change **if** you consider that a change to the output folder/file is a breaking change (it apparently wasn't). In the normal use case, which I guess is yours it shouldn't break, and this is a bug that needs fixing.

Basically,
* if:
```erlang
{log_root, "log"},
{crash_log, "crash.log"}
```
the file should be saved under
`log/crash.log`

* if:
```erlang
{log_root, "log"},
{crash_log, "log/crash.log"}
```
the file should be saved under
`log/crash.log`

I'll fix it but want to add a test for it too, this time around.

PR #541, opened 2021-02-18
Allow formatters to display stack traces on crash

lrascao:
Only the last stack frame (ie. the place that cause the crash) is currently being
shown in the message provided to the formatters, if you're interested in more
information you'll need to dig into the crash.log for it.
This is not ideal when using containerized environments with custom formatters,
we want as much as data as possible to be available in the console backend.

### DIFF
--- a/src/error_logger_lager_h.erl
+++ b/src/error_logger_lager_h.erl
@@ -391,12 +391,15 @@ format_crash_report(Report, Neighbours) ->
 
     {Class, Reason, Trace} = get_value(error_info, Report),
     {Md, ReasonStr} = format_reason_md({Reason, Trace}),
+    % `format_reason_md` only formats the last stack frame, we're also interested
+    % in making the rest of the stack available to formatters
+    TraceStr = format_trace(tl(Trace)),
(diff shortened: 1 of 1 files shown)


mrallen1:
IIRC we are only printing the last frame in the stack trace for safety reasons; that is, if you emit the entire stack trace you may be liable to OOM your VM if a gen_server with a giant 10MB state dies and gets printed over and over and over and over. So I am mildly concerned this patch opens a potentially unsafe log path.

Trust me, though I get why you want more information from a container environment, but I think it should be carefully considered.

lrascao:
That would indeed be bad, fortunately the crash report (to which this applies) does not carry the process state, that is left to the `=ERROR REPORT====` report.
However i understand your concern in terms of depth of trace, a very deeply nested function crash would fall into this issue. Would a (low default) configurable value to cap the maximum length for this list help in this regard? I'm thinking a depth of 3 or 4 would be enough to help sort out most crash origins.

mrallen1:
I would prefer to have a reasonable limit somewhere on the depth, yes, please.

PR #549, opened 2021-03-23
add support settings to specify multiple logging levels

dong50252409:
multiple log levels are supported for setting, like to
`{lager_file_backend,
    {file, "info.log"},
    **{level, [debug, error, notice]}**
}`

### DIFF
--- a/src/lager.erl
+++ b/src/lager.erl
@@ -440,17 +440,17 @@ get_sink_handler_status(Sink, Handler, Level) ->
     end.
 
 %% @doc Set the loglevel for a particular backend.
-set_loglevel(Handler, Level) when is_atom(Level) ->
+set_loglevel(Handler, Level) when is_atom(Level); is_list(Level) ->
     set_loglevel(?DEFAULT_SINK, Handler, undefined, Level).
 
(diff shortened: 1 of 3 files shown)


paulo-ferraz-oliveira:
Hm... how would that work? You'd only log those specific levels? I mean, because the way it is nowadays, `error` covers `notice` and `debug`, and also `info` and `warning`.

dong50252409:
The logging level can be specified this way in the configuration

>     {handlers, [
>         {lager_console_backend,[
>             {level, debug}          % Outputs all log levels to the console
>         ]},
>         {lager_file_backend, [
>             {file, "info.log"},     % The info.log file only records logs at the levels of DEBUG, INFO, NOTICE, AND WARNING
>             {level, [debug, info, notice, warning]}  
>         ]},
>         {lager_file_backend, [
>             {file, "error.log"},    % The error.log records contains all logging levels above ERROR
>             {level, error}
>         ]}
>     ]}.

You can also set the log level you want to view in the console like this

> lager:set_loglevel(lager_console_backend, [info, notice]).

Issue #558, opened 2021-06-11
'line' is a tuple on OTP-24


rlipscombe:
The parse transform assumes that the AST node location is the line number, and puts it into metadata as such. Since OTP-24, it's actually {Line, Col}.

See https://github.com/erlang-lager/lager/blob/3.9.2/src/lager_transform.erl#L148-L150

This survives all the way through to the default formatter, so I end up with logging that looks like the following:

```
2021-06-11 14:43:18.976 app@host [notice] <0.1763.0>@module:function:{168,5} Message
```

See that `{168,5}`? I was expecting `168`...

Seen with 3.9.2.

Something like the following ought to fix it (only lightly tested):

```
--- a/src/lager_transform.erl
+++ b/src/lager_transform.erl
@@ -141,13 +141,17 @@ do_transform(Line, SinkName, Severity, Arguments0) ->
 
 do_transform(Line, SinkName, Severity, Arguments0, Safety) ->
     SeverityAsInt=lager_util:level_to_num(Severity),
+    LineNum = case Line of
+        {L, _C} -> L;
+        L -> L
+    end,
     DefaultAttrs0 = {cons, Line, {tuple, Line, [
                                                 {atom, Line, module}, {atom, Line, get(module)}]},
                      {cons, Line, {tuple, Line, [
                                                  {atom, Line, function}, {atom, Line, get(function)}]},
                       {cons, Line, {tuple, Line, [
                                                   {atom, Line, line},
-                                                  {integer, Line, Line}]},
+                                                  {integer, Line, LineNum}]},
                        {cons, Line, {tuple, Line, [
                                                    {atom, Line, pid},
                                                    {call, Line, {atom, Line, pid_to_list}, [
```

rlipscombe:
Should probably also rename `Line` to `Loc`, but that would be a very noisy diff.

mrallen1:
Is there some reason why you want to throw away the column information? Or is it that it breaks some kind of reporting tool that you're using?

rlipscombe:
1. It looks unusual when formatted as a tuple. For consistency with most other tools, it should be `Line:Col`. For this, the metadata ought to have `line` and `col` attributes.
2. The column information doesn't _mean_ anything, as far as I can tell, because it's been run through a parse transform.

rlipscombe:
[...]

PR #560, opened 2021-06-11
Separate 'line' and 'col' metadata on OTP-24

rlipscombe:
See https://github.com/erlang-lager/lager/issues/558

### DIFF
--- a/src/lager_transform.erl
+++ b/src/lager_transform.erl
@@ -136,28 +136,41 @@ add_function_transforms(Line, DefaultAttrs, [{Atom, on_log, {Module, Function}}|
     add_function_transforms(Line, {cons, Line, NewFunction, DefaultAttrs}, Remainder).
 
 
+build_dynamic_attrs(Line) ->
+    {cons, Line, {tuple, Line, [
+                                {atom, Line, pid},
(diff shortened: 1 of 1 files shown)


src/lager_transform.erl
```diff
@@ -136,28 +136,41 @@ add_function_transforms(Line, DefaultAttrs, [{Atom, on_log, {Module, Function}}|
     add_function_transforms(Line, {cons, Line, NewFunction, DefaultAttrs}, Remainder).
 
 
+build_dynamic_attrs(Line) ->
```
rlipscombe, line 139: Should these go after `do_transform`? Wasn't paying enough attention to the current convention.

paulo-ferraz-oliveira:
Hi.

Any chance this moves forward? It seems @rlipscombe is waiting on a review 👍. To me the change seems Ok, since it keeps `line` as-was (non-breaking and restoring previous behaviour to consumer expectations) while adding `col` for more detailed consumption.

Cheers.

jadeallenx (approved):

jadeallenx:
Wow, so sorry about the extended delay. Thanks for the patch and the reminder.

PR #561, opened 2021-07-08
Add Windows + Erlang 24 to CI

paulo-ferraz-oliveira:
(moved away from `gleam-lang/setup-erlang` since it doesn't support Windows with Erlang 24)


paulo-ferraz-oliveira:
Ping.

mrallen1:
Thanks!

PR #568, opened 2021-11-19
Save state with new logfile handle when drop logs in shaper_expired

outddd:
There is a problem, when shaper expired and try drop logs, it can reopen log file, but doesn't save file handle. After this it reopen this file again, but first handle will never close, and old log file never delete from file system (until process dies)

Also try to escape edge-case bug: close reopened logfile with same name but new inode, when new file is exceed size limit.


jadeallenx (approved):
Thanks for the fix

